PR #1, opened 2022-05-20
Initial Commit

TomDijkema:
Initial Commit of repository.

Set up basic React folder structure and application.


TomDijkema:
I think I will need to remove all the node modules, because they will be included in the Docker image? So probably also add node modules to .gitignore.

PR #7, opened 2022-11-09
Feature digitalmediapage

TomDijkema:
This PR adds a view for the Digital Media Item that can be accessed through it's own unique URL. This page allows the user to see the digital media item and also to click through the other media items associated with the target specimen. Annotating media metadata is possible through the same annotation modal as on the specimen page, additionally a beginning for visual annotating of images has been added.

The annotation modal's code has been reworked a bit so it is easier to include this component in other pages as well. In the future it is probably more usefull to create a seperate folder for these kind of components instead of including it from the annotate folder that is nested in the specimen folder.

### DIFF
--- a/src/templates/digitalMedia/body/mediaAnnotations/MediaAnnotations.js
+++ b/src/templates/digitalMedia/body/mediaAnnotations/MediaAnnotations.js
@@ -1,0 +1,80 @@
+import { Row, Col } from 'react-bootstrap';
+
+/* Import Sources */
+import DigitalMediaFilterLayer from 'sources/digitalMediaFilterLayer';
+
+
+const MediaAnnotaions = (props) => {
+    const annotations = props.annotations;
+
+    if (annotations) {
+        if (Object.keys(annotations).length > 0) {
+            return (
+                <Row className="h-100">
+                    <Col>
+                        <Row>
+                            <Col className="digitalMedia_annotationsBlockHeader px-4 py-2 fw-bold">
+                                <Row>
+                                    <Col md={{ span: 7 }}>
+                                        Attribute
+                                    </Col>
+                                    <Col md={{ span: 5 }}>
+                                        Type
+                                    </Col>
+                                </Row>
+                            </Col>
+                        </Row>
+
+                        <Row>
+                            <Col className="digitalMedia_annotationsBlock">
+                                {Object.keys(annotations).map((group, _i) => {
+                                    const annotationGroup = annotations[group];
+
+                                    let annotationView = [];
+
+                                    for (const key in annotationGroup) {
+                                        const annotationObjects = annotationGroup[key];
+
+                                        for (const secondKey in annotationGroup[key]) {
+                                            const annotationObject = annotationObjects[secondKey];
+                                            const propertyObject = DigitalMediaFilterLayer[annotationObject['target']['indvProp']];
+
+                                            annotationView.push(
+                                                <Row>
+                                                    <Col className="digitalMedia_annotationsBlockRow px-4 py-2 border-b-1-primary-dark"
+                                                        onClick={() => props.ToggleModal(propertyObject, annotationObject['target']['indvProp'], annotationObject['motivation'])}
+                                                    >
+                                                        <Row key={secondKey}>
+                                                            <Col md={{ span: 7 }}>
+                                                                {`${annotationObject['target']['indvProp']} `}
+                                                            </Col>
+                                                            <Col md={{ span: 5 }}>
+                                                                {`${annotationObject['motivation']}`}
+                                                            </Col>
+                                                        </Row>
+                                                    </Col>
+                                                </Row>
+                                            );
+                                        }
+                                    }
+
+                                    return annotationView;
+                                })}
+                            </Col>
+                        </Row>
+                    </Col>
+                </Row>
+            );
+        } else {
+            return (
+                <Row>
+                    <Col className="digitalMedia_annotationsBlock px-4 py-2">
+                        No annotations yet
+                    </Col>
+                </Row>
+            );
+        }
+    }
+}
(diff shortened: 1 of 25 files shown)


src/templates/digitalMedia/body/mediaAnnotations/MediaAnnotations.js
```diff
@@ -0,0 +1,80 @@
...
+                                    <Col md={{ span: 5 }}>
+                                        Type
+                                    </Col>
+                                </Row>
+                            </Col>
+                        </Row>
+
+                        <Row>
+                            <Col className="digitalMedia_annotationsBlock">
+                                {Object.keys(annotations).map((group, _i) => {
```
samleeflang, line 30: I would place this map in a separate function

src/templates/digitalMedia/digitalMedia.scss
```diff
@@ -0,0 +1,147 @@
...
+}
+
+.digitalMedia_modalCloseButton {
+    cursor: pointer;
+}
+
+
+/* Media Modal Image Annotate */
+
+.hAjSKD {
```
samleeflang, line 140: Are these names correct, are they abbreviations or just random?

src/templates/digitalMedia/digitalMedia.scss
```diff
@@ -0,0 +1,147 @@
...
+    line-height: 40px;
+    font-size: 20px;
+}
+
+.digitalMedia_mediaBodyImages {
+    height: Calc(100% - 40px) !important;
+}
+
+.digitalMedia_mediaBodyTab {
+    height: 40px;
```
samleeflang, line 22: Still a lot of fixed heights based on px. How will that scale if on smaller/bigger screens?

src/templates/home/body/recentAnnotations/RecentAnnotations.js
```diff
@@ -45,8 +45,18 @@ const RecentAnnotations = () => {
                                     <Col md={{ span: 12 }}>
                                         <Row>
                                             {recentAnnotations.map((annotation, i) => {
-                                                if (!annotation['specimen']['specimenName']) {
-                                                    annotation['specimen']['specimenName'] = 'Undefined name';
+                                                /* Temporary solution */
```
samleeflang, line 48: Create a list for the endpoints you would like. This seems something to be fixed in a endpoint where we can add a parameter to return only specimen annotations or only media annotations.

src/index.js
```diff
@@ -7,7 +7,7 @@ import axios from "axios";
 import "react-datepicker/dist/react-datepicker.css";
 
 
-axios.defaults.baseURL = "http://localhost:3000/api/v1";
+axios.defaults.baseURL = process.env.REACT_APP_AXIOS_URL;
```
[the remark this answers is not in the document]
TomDijkema: Hier werkt het idd wel, dan pas ik de env na merge in de main branch aan naar de sandbox omgeving. Kon alleen nog geen oplossing vinden voor de proxy in package.json, daar kan je zover ik kon zien niet gebruik maken van .env variabelen. Even kijken hoe dat het beste gedaan kan worden

src/templates/digitalMedia/digitalMedia.scss
TomDijkema: Yeah, they are classes used by the image annotation library, found them by inspecting cause I needed an inner element to stretch to the correct width and height which I could not achieve by simply passing classes onto the element, will give add a more clearer description in the CSS file

src/templates/digitalMedia/body/mediaAnnotations/MediaAnnotations.js
TomDijkema: Removed map function from jsx and moved to functional code, prepares array and then renders the components in the array in the view

src/templates/home/body/recentAnnotations/RecentAnnotations.js
TomDijkema: Will create a list

TomDijkema: Get Annotation calls (all, recent, etc.): join op de target (Specimen/Media) om ipv een id de naam of iets anders van het target te tonen
Get Annotation calls (all, recent, etc.): paginate op calls, pageSize en pageNum parameters
Search Specimens call: total count van alle gevonden specimens
Get Media calls (all, recent): join op de target Specimen om ipv een id de naam te tonen

src/templates/digitalMedia/digitalMedia.scss
TomDijkema: This is an probable issue, also in connection with the whole mobile capacity of UCAS, which is minimal in the current situation

src/templates/home/body/recentAnnotations/RecentAnnotations.js
samleeflang: Ok heb hier een user stories voor aangemaakt in Jira zodat we ze niet kwijt raken.
https://naturalis.atlassian.net/browse/DD-39

src/templates/digitalMedia/digitalMedia.scss
samleeflang: Ok something to keep in mind but which we cannot solve in this PR.

src/templates/digitalMedia/digitalMedia.scss
samleeflang: Ok maybe leave an additional comment in the code with this so future dev's will understand

samleeflang: Ah I now see the change, ok should be enough

samleeflang (approved):
Approved

PR #10, opened 2023-01-25
Add simple nav dropdown for profile

TomDijkema:
Commit adds a simple dropdown in the nav bar for accessing the user profile or to log-out when logged in. 'Simple' because of upcoming style changes from the UX team, used standard Bootstrap component

### DIFF
--- a/src/keycloak/Keycloak.js
+++ b/src/keycloak/Keycloak.js
@@ -38,7 +38,7 @@ const InitKeyCloak = (callback) => {
 
 const doLogin = keycloak.login;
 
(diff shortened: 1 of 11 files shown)


TomDijkema:
! After Merge, login needs to be checked since it is constantly kicking out the active Keycloak session on the local instance !

TomDijkema:
> ! After Merge, login needs to be checked since it is constantly kicking out the active Keycloak session on the local instance !

Probably not necessary, the login problems have to do with Firefox handling third-party cookies on a local instance

samleeflang (approved):

PR #11, opened 2023-02-16
Type script update

TomDijkema:
The Type Script Update:

This Pull Request updates the whole application to make use of Type Script in order to make the Javascript code strongly typed and help with type indications in code. Basically it becomes to indicate types of variables when programming and should help future contributors understand the code.

For the implementation of Type Script a general file containing all common types has been created. Types like: User, Specimen, DigitalMedia and Annotation are imported frequently and form the basis for a lot of the content. Other, less significant types, like element props, have been integrated into the code itself.

Apart from Type Script, has every file  of the application been revised in order to make some quality of life changes with newer, more efficient learnt techniques. For example: a Redux Store has been integrated into the application which can hold primary scoped states which relieves a lot of components from setting loads of props to children and have the advantage to be callable and mutable from every location. Example: when landing on the Specimen Page, it will look if there is an active specimen in the specimen slice of the store, if not it will grab the specimen from the database and store it, able to persist it through navigating to other pages. Only a refresh causes the Redux store to be reconfigured.

All API functions have been transformed to asynchronous functions which return a promise. The promise can be fetched by the component that called on the API function to extract the returned data. This method gets rid of all the callbacks in the code which makes it more fluent and readable from top to bottom. Additionally, all functions have been transformed to arrow functions (example: function functionName() {} now is: const functionName = () => {}.

Introduced more comments in the code to explain what is happening. Also to split up the import statements by their respective source.

Changed the templates' folder name to: components. Changed the order of components recurring in folders: instead of using a 'body' in between the parent and child components, primary child components are now imported into the parent component which also includes most overarching functionality. Otherwise, specific functionality for components has been tried to keep to the child components them self.

The Annotate Modal has been refined even more to the point it can almost be imported at any point of the code, as long as there is a valid object (specimen, digitalMedia) to be annotated.

This update still needs some patches to resolve minor issues.


src/api/annotate/InsertAnnotation.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Annotation, AnnotationTemplate } from 'global/Types';
+
+
+const InsertAnnotation = async (annotationRecord: AnnotationTemplate, token?: string) => {
+    if (annotationRecord && token) {
+        let annotation = <Annotation>{};
+
+        const endPoint = '/annotations';
+
+        await axios({
```
samleeflang, line 14: I think that the .then and the await do a similar thing. Are you sure this is correct. If I look at: https://rapidapi.com/guides/axios-async-await It is either the .then or the await

src/api/specimen/GetRecentSpecimens.ts
```diff
@@ -0,0 +1,26 @@
...
+/* Import Types */
+import { Specimen } from "global/Types";
+
+
+const GetRecentSpecimens = async () => {
+    let recentSpecimens = <Specimen[]>[];
+
+    const endPoint = "/specimens"
+
+    axios({
```
samleeflang, line 13: Here you don't use the await I think this is how it should look

src/api/specimen/GetSpecimenAnnotations.ts
```diff
@@ -0,0 +1,37 @@
...
+        const endPoint = `specimens/${handle}/annotations`;
+
+        await axios({
+            method: "get",
+            url: endPoint,
+            responseType: 'json',
+        }).then((result) => {
+            const annotations: Annotation[] = result.data;
+
+            /* Refactor Annotations object */
```
samleeflang, line 21: Is this still correct? Add a backlog item for this

src/api/user/GetUserAnnotations.ts
```diff
@@ -0,0 +1,43 @@
...
+            method: "get",
+            url: endPoint,
+            responseType: 'json',
+            headers: {
+                'Authorization': `Bearer ${token}`,
+            },
+        }).then((result) => {
+            userAnnotations = result.data;
+
+            /* Temporary solution to include Specimen data */
```
samleeflang, line 27: Add a backlog item for this

src/components/annotate/modal/ValueField.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    if (targetType === 'digital_specimen') {
+        filteredAnnotation = annotationFilterLayer[property];
+    } else if (targetType === 'digital_media') {
+        filteredAnnotation = digitalMediaFilterLayer[property];
+    }
+
+    /* Function for tracking start date if property is time related */
+    
+    /* Future Development: implement date field
```
samleeflang, line 37: Ah a nicely commented out piece. Is this still relevant or can it be deleted?

src/components/digitalMedia/components/annotate/ObservationModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+const ObservationModal = (props: Props) => {
+    const { observationModalToggle, ToggleObservationModal } = props;
+
+    /* Base Variables */
+    const digitalMedia = useAppSelector(getDigitalMedia);
+
+    /* Functionality for handling new, changed or removed Observation Annotations */
+    const [observationAnnotations, setObservationAnnotations] = useState<IAnnotation[]>(
+        []
+        // Future Development: implement DB annotations: digitalMediaAnnotations.observation ? digitalMediaAnnotations.observation : []
```
samleeflang, line 31: Make an userstory out of this?

src/components/digitalMedia/components/annotate/ObservationModal.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <Row className="h-100 justify-content-center">
+                <Col className="h-100">
+                    <Row className="h-100">
+                        <Col md={{ span: 10 }} className="bg-white pe-0 h-100 position-relative">
+                            <Annotation
+                                src={digitalMedia.mediaUrl}
+                                alt={digitalMedia.id}
+                                onAnnotationsUpdate={(annotations) => setObservationAnnotations(annotations)}
+                                annotations={observationAnnotations}
+                                // Future Development: implement DB annotations: annotations={observationAnnotations}
```
samleeflang, line 45: Same

src/components/digitalMedia/components/digitalMedia/DigitalMediaProperties.tsx
```diff
@@ -0,0 +1,56 @@
...
+                                    </Col>
+                                </Row>
+                            </Col>
+                        </Row>
+                    );
+                })}
+
+                <> </>
+
+                {/* <Row>
```
samleeflang, line 44: Commented out code

src/components/home/components/Mids.tsx
```diff
@@ -22,25 +23,31 @@ const Mids = () => {
...
+                                    <Col md={{ span: 3 }} key={level}>
                                         <Row>
                                             <Col md={{ span: 10 }}>
                                                 <div>
                                                     <div className="home_midsTitleBlock py-2 bg-primary-dark fw-bold text-white text-center">
                                                         MIDS level {level}
                                                     </div>
                                                     <div className="px-4 py-3 br-bl br-br border-2-primary-dark">
-                                                        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+                                                        Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore
```
samleeflang, line 48: Maybe add a task on the backlog to fix the text here. See if we can get Elspeth or someone to provide some text on MIDS

src/components/profile/components/annotate/UserProfileAnnotation.tsx
```diff
@@ -1,19 +1,29 @@
...
+    const annotationTypeMapping: Dict = {
         commenting: 'Comment',
         adding: 'Addition',
         correcting: 'Correction',
         quality_flagging: 'Quality Flag',
         linking: 'Relationship/Link'
     }
 
-    const isoDate = new Date(Date.parse(annotation['created']));
+    const isoDate = new Date(annotation.created);
```
samleeflang, line 26: This looks sooo much better :partying_face:

src/components/specimen/Specimen.tsx
```diff
@@ -0,0 +1,208 @@
...
+    const CheckAnnotations = (specimen: SpecimenType) => {
+        GetSpecimenAnnotations(specimen.id).then((annotations) => {
+            if (annotations) {
+                dispatch(setSpecimenAnnotations(annotations));
+            }
+        });
+    }
+
+    /* Function for a combined check on Specimen Digital Media, Annotations and Versions */
+    const CheckDetails = (specimen: SpecimenType) => {
```
samleeflang, line 105: We should really make a single call out of this. At least for media and specimen as in the next version annotations will be behind a tab (so we call annotations when someone clicks that tab). I will make a userstory for this

samleeflang: https://naturalis.atlassian.net/browse/DD-255

src/reportWebVitals.ts
```diff
@@ -0,0 +1,15 @@
+import { ReportHandler } from 'web-vitals';
+
+const reportWebVitals = (onPerfEntry?: ReportHandler) => {
```
samleeflang, line 3: What does htis do?

samleeflang (commented):
Couple small comments/questions

public/manifest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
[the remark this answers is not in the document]
TomDijkema: Done

src/api/annotate/InsertAnnotation.ts
TomDijkema: It is not quite the same because await really waits for the call to complete before anything in the function can be executed, .then does execute when the result is reeled in, but the other stuff in the function will continue. Chose await over .then for the API calls because of the waiting, think this is important. Removed .then for all api calls and introduced try/catch for checking if the fetch was successful or not.

src/api/specimen/GetRecentSpecimens.ts
TomDijkema: It is not quite the same because await really waits for the call to complete before anything in the function can be executed, .then does execute when the result is reeled in, but the other stuff in the function will continue. Chose await over .then for the API calls because of the waiting, think this is important. Removed .then for all api calls and introduced try/catch for checking if the fetch was successful or not.

src/api/specimen/GetSpecimenAnnotations.ts
TomDijkema: I think this is still preferred, by splitting it per property I can easily access the annotations of for example: the scientific name, in the view. That is why I created a separate function for this, only for the Specimen page since the other stuff just lists all the annotations.

src/api/user/GetUserAnnotations.ts
TomDijkema: Soulaine already had a look at this and said it was challenging, but will add it as a separate user story

TomDijkema: https://naturalis.atlassian.net/browse/DD-256?atlOrigin=eyJpIjoiZDlhZjliNGExNDkwNDBhYjhiODg0NTE4OTg5MzUwMmMiLCJwIjoiaiJ9

src/components/annotate/modal/ValueField.tsx
TomDijkema: I think this is still relevant because there are more types than just string and dates, think of select fields and booleans, point for discussion I guess.

https://naturalis.atlassian.net/browse/DD-257?atlOrigin=eyJpIjoiOGZmMDNmYTUzMWVlNDgxYWI0ZGEyMTM5ZDQ5NDJjMmEiLCJwIjoiaiJ9

src/components/digitalMedia/components/annotate/ObservationModal.tsx
TomDijkema: https://naturalis.atlassian.net/browse/DD-258?atlOrigin=eyJpIjoiMmJjMWU1ZjY5ZTFjNDA4YTliZmQ0M2U2NWZkMGQyNWEiLCJwIjoiaiJ9

src/api/user/GetUserAnnotations.ts
samleeflang: Think it got a lot easier not we removed all the versions from the database. Should be easy now

src/components/digitalMedia/components/digitalMedia/DigitalMediaProperties.tsx
TomDijkema: Removed

src/reportWebVitals.ts
TomDijkema: This is default React functionality that can be used to measure the performance and user experience of the application. Maybe useful during development, will probably never be active when deployed: https://create-react-app.dev/docs/measuring-performance/

samleeflang: Ok sounds good!

samleeflang (approved):
Looks good!

PR #12, opened 2023-02-27
Add id card

TomDijkema:
Pull Request will add the ID Card of a Specimen to the Specimen Page. Two institution logos are included to test the visual appearance.


src/components/specimen/components/idCard/IDCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+import MidsBar from './MidsBar';
+
+
+const IDCard = () => {
+    /* Base variables */
+    const specimen = useAppSelector(getSpecimen);
+
+    const logo = (ror: string) => {
+        try {
+            return require(`../../../../webroot/img/organizationLogo/${ror}.png`);
```
samleeflang, line 22: Does this mean we have to have all logo's in this folder?

samleeflang (approved):
Maybe make a backlog item to include all logo's in FE. Although they need to be small otherwise everything will be slow... Might need another solution there, maybe get it when the specimen loads (from ROR?) and then store in local cache or something.

src/components/specimen/components/idCard/IDCard.tsx
TomDijkema: https://naturalis.atlassian.net/browse/DD-265?atlOrigin=eyJpIjoiYjA4NDg4NTcwYzMzNDRkN2I3ZWRlNmQ1MWFiYTY3MjMiLCJwIjoiaiJ9

PR #13, opened 2023-02-27
Specimen harmonized view

TomDijkema:


src/components/annotate/modal/ValueField.tsx
```diff
@@ -22,12 +22,14 @@ const ValueField = (props: Props) => {
...
 
     /* Base variables */
-    const annotationFilterLayer: Dict = AnnotationFilterLayer;
+    const harmonisedAttributes: Dict = HarmonisedAttributes;
     const digitalMediaFilterLayer: Dict = DigitalMediaFilterLayer;
     let filteredAnnotation;
 
     if (targetType === 'digital_specimen') {
-        filteredAnnotation = annotationFilterLayer[property];
+        console.log(property);
```
samleeflang, line 30: We can probably remove the console.log right?

src/components/specimen/components/contentBlock/OriginalData.tsx
```diff
@@ -0,0 +1,109 @@
...
+                                        highlightOnHover
+                                    />
+                                </Card>
+                            </Col>
+                            <Col md={{ span: 3 }}>
+                                <Row>
+                                    <Col>
+                                        <Row>
+                                            <Col>
+                                                Harvested from:
```
samleeflang, line 78: Not sure about this part.
As it currently is it is incorrect as this data doesn't come from GBIF.
If we want to keep the Harvested from we should display the URL where we collected the information.
We have that in the sourceSystem tabel.
I would remove GBIF from here and make fill this with the SourceSystemURL, I think the endpoint is already there

samleeflang: https://sandbox.dissco.tech/api/v1/source-systems/prefix/suffix

src/components/specimen/components/contentBlock/SpecimenOverview.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+
+const SpecimenOverview = () => {
+    /* Base variables */
+    const specimen = useAppSelector(getSpecimen);
+
+    /* Check for Organization logo */
+    const logo = (ror: string) => {
+        try {
+            return require(`../../../../webroot/img/organizationLogo/${ror}.png`);
```
samleeflang, line 19: Make sure that everything will function when the logo is not there. For some smaller organisations we might never get a logo

src/components/specimen/components/contentBlock/SpecimenOverview.tsx
```diff
@@ -0,0 +1,155 @@
...
+                    <Col md={{ span: 7 }} className="pb-2">
+                        <Card className="position-relative h-100">
+                            <Card.Body>
+                                <Card.Title>
+                                    Location
+                                </Card.Title>
+
+                                <Row className="mt-3">
+                                    <Col>
+                                        Country: Netherlands
```
samleeflang, line 39: I think it is better to keep dummy values in the fields we haven't got harmonised data for. Just so it is very clear to anyone coming to the page that this still needs to be done. Maybe something like "In Progress" Or "Coming Soon" something to make clear that it is not actual data.
Say we ingest a bunch from Paris and the for some reason check our site we don't want them to think we messed up their data and it is now in the Netherlands or something

src/sources/hamonisedAttributes.json
```diff
@@ -0,0 +1,39 @@
+{
+    "physicalSpecimenId": {
+        "displayName": "Physical Specimen ID"
+    },
+    "ods:physicalSpecimenIdType": {
+        "displayName": "Physical Specimen ID Type"
+    },
+    "ods:organizationId": {
+        "displayName": "Organization",
+        "annotate": {
```
samleeflang, line 10: What does the `annotate` do?

samleeflang (commented):
Couple of small comments

src/components/specimen/components/contentBlock/ContentBlock.tsx
```diff
@@ -0,0 +1,148 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faChevronDown } from '@fortawesome/free-solid-svg-icons'
+
+/* Import Components */
+import SpecimenOverview from './SpecimenOverview';
+import OriginalData from './OriginalData';
+import AnnotationsOverview from './AnnotationsOverview';
+import DigitalMedia from './DigitalMedia';
+import MIDSOverview from './MIDSOverview';
+// import SpecimenProperties from './components/specimen/SpecimenProperties';
```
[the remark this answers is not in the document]
TomDijkema: Removed

src/components/annotate/modal/ValueField.tsx
TomDijkema: Removed

src/components/specimen/components/contentBlock/OriginalData.tsx
TomDijkema: Created function to handle API call to Source Systems and replaced static content with actual representation of Source System

src/components/specimen/components/contentBlock/SpecimenOverview.tsx
TomDijkema: Yes, that is why I use the try/catch combination, if the file does not exist, it does not break the code at the moment. Personally I would prefer to still do it a bit different because the require/import is pretty ugly due to all the ../../../, something to look at in the future when we collect more logos

src/components/specimen/components/contentBlock/SpecimenOverview.tsx
TomDijkema: Yes that sounds like something that could happen, changed all the dummy fields to: 'Coming Soon'.

src/sources/hamonisedAttributes.json
TomDijkema: The idea was to include configuration properties in this object for when annotating this harmonised property.
For example: the organizationId field will be bound to the organizations participating in DiSSCo, via the JSON or some other source in the future, I can determine that the form field type when annotating should be a select dropdown containing all the organizations as options. This would also probably only be applicable when using 'adding' or 'correcting' as a motivation.

src/components/specimen/components/contentBlock/OriginalData.tsx
samleeflang: Nice!

src/sources/hamonisedAttributes.json
samleeflang: Ok yes good to think ahead :+1:

samleeflang (approved):

PR #14, opened 2023-03-02
Internationalization

TomDijkema:
Pull Request adds the i18n-next library to setup multi lingual support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "dotenv-cli": "^6.0.0",
     "formik": "^2.2.9",
     "html-react-parser": "^3.0.4",
+    "i18next": "^22.4.10",
     "keycloak-js": "^18.0.0",
     "moment": "^2.29.4",
     "react": "^18.2.0",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -23,6 +23,7 @@
     "dotenv-cli": "^6.0.0",
     "formik": "^2.2.9",
     "html-react-parser": "^3.0.4",
+    "i18next": "^22.4.10",
```
samleeflang, line 26: Do you need them both? `i18next` and `react-i18next`

samleeflang (approved):
Already spoken about approach. Looks good, just one small quesion

package.json
TomDijkema: I used both. According to their documentation i18n-next is the core library that sets up the language support. The extension for React adds a hook method so the translate function (t) can easily be called on in components to use the translate functionality.

TomDijkema: Probably could do without, but should be more optimized.

samleeflang: Ok clear

PR #17, opened 2023-04-05
Generalise Filtering

TomDijkema:
Commit generalizes the filtering by using the new Aggregation endpoint. Filters are now rendered based upon the returned values from this endpoint (when accepted in the front-end) and displayed in the filters menu. Each option received a checkbox with the corresponding values and aggregation numbers.

The Multi Select Component for rendering the filters has been cleaned up so it accepts all kinds of filter names and values and spits out checkboxes with the ability to add these to the Search Filters, as well as to remove them.

Adds:
- Filters based upon the Aggregation endpoint using the Multi Select component (checkboxes)
- Aggregation numbers on selectable/selected filters

Modifies:
- Multi Select Component to be general and accept all kinds of filter names and values
- Search bar to use the 'q' search filter for the free text search
- Organisation Search on the Home Page remains the use of the ROR to filter on organisations, requires update from the API

Bug Fixes:
- Resolve source system not showing up on Specimen page due to wrong route call of endpoint (needs to call /orchestrator and not /api/v1)


src/api/sourceSystem/GetSourceSystem.ts
```diff
@@ -9,12 +9,13 @@ const GetSourceSystem = async (handle: string) => {
...
         let sourceSystem = <SourceSystem>{};
 
-        let endPoint: string = `source-systems/${handle}`;
+        let endPoint: string = `source-system/${handle}`;
 
         try {
             const result = await axios({
                 method: "get",
                 url: endPoint,
+                baseURL: 'https://sandbox.dissco.tech/orchestrator',
```
samleeflang, line 18: Oh crap just changes this. Is this going to break on CORS, probably (sorry :crying_cat_face: )

TomDijkema: Changed route

samleeflang (approved):

PR #18, opened 2023-04-11
Add Harmonized Fields

TomDijkema:
Commit adds harmonized fields to the specimen overview tab on the Specimen Page. These harmonized fields should demonstrate how we can display values conform different categories like: location, organisation and collection. Harmonized fields are annotatable.

Commit also adds a Leaflet map that can be generated based on the coordinates of the specimen. A placeholder will be shown if the latitude or longitude is missing.

Adds:
- Couple of harmonized fields to display how we can handle this (annotatable)
- Leaflet (OpenStreet) map based upon the coordinates of the specimen

Bug fixes:
- Cannot annotate physical specimen id due to its missing prefix 'ods:'  in the code

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,6 +30,7 @@
         "html-react-parser": "^3.0.4",
         "i18next": "^22.4.10",
         "keycloak-js": "^18.0.0",
+        "leaflet": "^1.9.3",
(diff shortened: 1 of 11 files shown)


samleeflang (approved):
Looks good, no comments!

PR #21, opened 2023-04-19
Overall enhancements

TomDijkema:
PR comes with overall enhancements based upon smaller user stories:
- Fixes the pagination on the Search Page
- Reworks the Digital Media Tab on the Specimen Page
- Reworks the Header Component
- Reworks the Footer Component
- Adds feedback option to the Footer
- Rebrands the application to: DiSSCoVer
- Fixes several small bugs

Keep in mind the rebranding of the application requires a clean restart of the Kubernetes pod since the metadata containing the name UCAS have been replaced with DiSSCoVer.
**The repository in AWS has not yet been changed**.


src/components/general/footer/Footer.tsx
```diff
@@ -1,93 +1,50 @@
...
+import classNames from "classnames";
 import { Container, Row, Col } from "react-bootstrap";
 
 /* Import Styles */
-import './footer.scss';
-
-/* Import Icons */
-import { FontAwesomeIcon } from "@fortawesome/react-fontawesome";
-import { faChevronUp } from "@fortawesome/free-solid-svg-icons";
+import styles from './footer.module.scss';
```
samleeflang, line 7: Shouldn't this also be double quotes?

src/components/general/header/components/Login.tsx
```diff
@@ -2,18 +2,17 @@
 import KeycloakService from 'keycloak/Keycloak';
 import { Row, Col } from 'react-bootstrap';
```
samleeflang, line 3: I see the mix between single and double quotes, not sure if it needs fixing, maybe not in this PR.

src/components/search/Search.tsx
```diff
@@ -38,6 +38,9 @@ const Search = () => {
     /* Base variables */
     const searchResults = useAppSelector(getSearchResults);
     const searchSpecimen = useAppSelector(getSearchSpecimen);
+    const pageSize = 25;
```
samleeflang, line 41: We could also set 25 as default value, currently default is 10

samleeflang (approved):

src/components/general/footer/Footer.tsx
TomDijkema: In principle, it does not matter, but overall I decided to use single quotes for import statements.
Some of the older files still got the double quote imports, will replace any I find with single quotes.

src/components/general/header/components/Login.tsx
TomDijkema: My intent is to use single quotes with import statements, will replace any old imports with double quotes I encounter

src/components/search/Search.tsx
TomDijkema: Yes, that would be good, this number is specially for the paginator component, so not dependent on the API (but used in the params with the call). I think 25 is a good number, not too big, not too small, otherwise we could increase or decrease it a little

PR #22, opened 2023-04-24
Rework profile

TomDijkema:
PR completely reworks the User Profile to be more inline with the design. Also adds links to the ORCID and Bionomia accounts of the user, as well as pagination to the user annotations table.

Adds:
- ORCID and Bionomia links to Profile
- Pagination to user annotations table

Modifies:
- Overall lay-out and appearance of the user profile

Bug fixes:
- Paginator now is based upon variables from the source file (containing the table) which should prevent double looping as well as entering page numbers that contain no data


src/components/general/paginator/Paginator.tsx
```diff
@@ -1,35 +1,30 @@
...
-import { useEffect, useState } from 'react';
-import { Pagination } from 'react-bootstrap';
+import { Pagination } from "react-bootstrap";
 
 
 /* Props Typing */
 interface Props {
     pageSize: number,
     pageNumber: number,
+    paginationRange: number[],
```
samleeflang, line 9: Does this fix the issue with the pagination? I noticed that if you move to the end then it displays no more results. So you basically go beyond the items that are there. I also think we can show at least the next and previous page. Based on the `next` field in the json `links` you can see if there is a next page and display it.

samleeflang (approved):

src/components/general/paginator/Paginator.tsx
TomDijkema: Yes, I modified the Paginator component so it will not longer continue going to pages without data.
Really? That is so much easier than the method I used in this commit!
Committed a new commit to fix this based upon the `links` data, omits the pagination range.

PR #27, opened 2023-05-12
Rework filters

TomDijkema:
PR reworks filters on Search page according to the design. Aside from looks, the behavior of the multi select filters have changed a bit to be more efficient plus the addition of date and free text filter fields (not yet functional in search). Filter tags, above the results table, also have been added. These can be used to freely look and remove filters, also when the filters menu itself is hidden. If the amount of filter tags exceeds the width and height of the parent block, a dropdown will become actionable to view all the filter tags (this will hover over the results table).

Adds:
- Improved filter functionality
- Date and free text filter fields (not yet functional in search)
- Filter tags above results table
- Dropdown for filter tags if length exceeds parent block's width (height)

Resolves:
- Indexing bug that caused multi select filter checkboxes to deactivate the wrong index, hence the wrong filter


src/components/search/components/searchMenu/SearchFilters.tsx
```diff
@@ -0,0 +1,204 @@
...
+                                        />
+                                    </Col>
+                                </Row>
+                                <Row>
+                                    <Col>
+                                        {Object.keys(filtersList).map((filterKey) => {
+                                            const filter = filtersList[filterKey];
+
+                                            /* Check kind of filter */
+                                            switch (filter.filterType) {
```
samleeflang, line 162: I don't see mids in this switch is that correct?

samleeflang (approved):

src/components/search/components/searchMenu/SearchFilters.tsx
TomDijkema: That's correct, I treated MIDS essentially as a multi select filter, but with a different template. So the MIDS template uses the same functionality as the default Multi Select template. If I would have put MIDS in this select, you would get a lot of repetitive code in the MIDS file that already is in the Multi Select one. That's why I decided to put MIDS a level lower, looking at the type in the Multi Select component to render the default or MIDS template.

PR #30, opened 2023-05-23
Add Compare Mode

TomDijkema:
PR adds Compare Mode to the Search Page, allowing the user to select up to three Specimen to compare. Upon selecting and confirming, a new page shows up displaying the selected Specimens in comparison mode. This url is also approachable using the route and the ids from the specimens as params (like: search/compare?ds=1234&ds=4321).

Adds:
- Compare Mode
- Select Specimens in Data Table
- New Page and route: /search/compare
- Compare Cards (similar to ID Card) on Compare Page for Specimen comparison

Behavior changes:
- ID Card will always hide upon activating Compare Mode
- Filters will remain in place when activating Compare Mode


samleeflang (approved):
Nothing to complain. Looking good!

PR #31, opened 2023-05-26
Mvp roundup

TomDijkema:
PR for MVP roundup checks

### DIFF
--- a/src/components/home/components/specimenTypes/SpecimenTypeFilters.tsx
+++ b/src/components/home/components/specimenTypes/SpecimenTypeFilters.tsx
@@ -1,5 +1,5 @@
 /* Import Dependencies */
-import { useEffect, useState } from 'react';
+import { useEffect, useState, useRef } from 'react';
 import { useNavigate } from 'react-router-dom';
 import { Formik, Form, Field } from 'formik';
 import CountUp from 'react-countup';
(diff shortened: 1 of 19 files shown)


src/components/home/components/specimenTypes/SpecimenTypeFilters.tsx
```diff
@@ -50,6 +51,9 @@ const SpecimenTypeFilters = () => {
         });
     }, []);
 
+    /* Calculate total amount of Specimens */
+    const totalSpecimenCount = (Object.values(disciplines).reduce((accumulator, currentValue) => accumulator + currentValue));
```
samleeflang, line 55: I believe there is a total value in the API Response as well right? You could use that, so you don't have to calculate in the frontend

samleeflang: "meta": {
        "totalRecords": 2241740
    }

samleeflang (commented):
Small thing about the totals on homepage. This can be take from the response, instead of calculated

src/components/home/components/specimenTypes/SpecimenTypeFilters.tsx
TomDijkema: Did not see that, implemented the total records returned from the API instead of calculating it in the front-end

samleeflang (approved):

PR #36, opened 2023-07-17
HomePage Responsiveness - Rework Filters

TomDijkema:
PR conceptualizes how responsiveness could be implemented using simple Bootstrap and Flex classes (needs further discussion and refinement).
PR also reworks filters on Search Page to be inline with new design. Active filters are moved into the filters menu when visible, still above the table when collapsed. Search bar has been readded above the table and is visible at all times. The query parameter will not be shown as one of the chosen filters as this is already visible in the search bar.

Adds:
- Search Bar above search table, visible at all times
- Added a source file to be used instead of a loose variable to declare the search filters

Modifies:
- Homepage components and classes to make it responsive for tablet sized devices (needs further refinement)
- Active filters to be only visible when the filters menu is collapsed, otherwise they are displayed in the filters menu itself.
- Fix some aligning issues with the filters menu


samleeflang (approved):
Looks good!

PR #37, opened 2023-07-18
Annotations side panel

TomDijkema:
PR adds the annotation side panel. This side panel will replace the old modal for managing (user) annotations. The side panel can be triggered for all annotations, or for a specific property of a specimen, which immediately sets the target field. A user can add annotations by pressing the 'add' button, which will trigger the annotation form. Select a motivation and target property (if not already), and insert the required values into the form fields. Adding an annotation directly adds it to the overview. Similar, annotations can be patched by clicking on the pen icon, this opens the form again, but with prefilled values to edit. At last, annotations can be removed by clicking on the trash can icon. It will ask the user for confirmation, where after the annotation will be removed.

Adds:
- Annotations side panel
- Add user annotations
- Edit annotations
- Delete annotations

Removes:
- Old annotations modal


src/components/annotate/sidePanel/Annotation.tsx
```diff
@@ -0,0 +1,200 @@
...
+        /* Reset hightlight annotation id */
+        dispatch(setHighlightAnnotationId(''));
+
+        /* Highlight 1 */
+        annotationDiv.classList.add(`${styles.highlight}`);
+
+        setTimeout(() => {
+            annotationDiv.classList.remove(`${styles.highlight}`);
+
+            setTimeout(() => { Second() }, 500);
```
samleeflang, line 73: Is the second timeout needed?

src/components/annotate/sidePanel/Annotation.tsx
```diff
@@ -0,0 +1,200 @@
...
+                            />
+                        </Col>
+
+                        {KeycloakService.GetSubject() === annotation.creator &&
+                            <Col className="col-md-auto">
+                                <Row>
+                                    <Col className="pe-0">
+                                        <FontAwesomeIcon icon={faPen}
+                                            className="c-pointer c-primary"
+                                            onClick={() => dispatch(setEditAnnotation(annotation))}
```
samleeflang, line 165: The user should only be able to update/delete their own annotation. Do we check this in the FE?

TomDijkema: Yes. The edit and delete icons are only generated when the user is logged in and if the creator id of the annotation is the same as the user's keycloak id.

image

src/components/annotate/sidePanel/Annotation.tsx
TomDijkema: No, was already thinking about this. Replaced with an interval and timeout, which also loops the highlight three times in total.

samleeflang: :+1:

src/components/annotate/sidePanel/Annotation.tsx
samleeflang: Check, thanks for explaining

samleeflang (approved):

PR #39, opened 2023-07-24
Add Generic Annotation to Specimen

TomDijkema:
Commit adds the possibility to add a generic annotation to a specimen. It also better indicates the difference between property bound annotations and the generic ones. The back button (chevron) on the top of the side panel now acts different, going back to the overview first when editing or viewing an individual property.

Adds:
- Add generic annotations
- Back button functionality for the top chevron in the side panel


southeo (approved):
:100:

PR #41, opened 2023-07-27
Add Annotation Side Panel to Digital Media

TomDijkema:
Commit adds Annotation Side Panel to Digital Media page. All annotation features are included: creating, editing and deleting annotations. Actually, this functionality is supplied by the side panel itself, and is not repeatably used. Only the update functions for the original Digital Media annotations are similar to the Specimen counterpart and could be generalised in the future.

Adds:
- Annotations Side Panel to Digital Media page

Modifies:
- Digital Media ID Card's properties to not trigger the side panel, these need refinement and decide on whether these are harmonised properties or not
- Aligning of the components on the Digital Media page to fit the side panel, and be relative to the Specimen Page


southeo (approved):
:rock: :sauropod: :fish:

TomDijkema:
Duplication was intended, did not want to edit the Sonar Cloud rules too much, so force merged this one.

PR #42, opened 2023-07-31
Add refresh button to Side Panel in order to refresh Annotations

TomDijkema:
Commit adds a refresh button to the Side Panel that can be used to refresh the annotations within it. This method is dependent upon the parent component passing it on, otherwise the button will not be displayed. The Specimen of Digital Media component will trigger a refresh call to the API and provide the annotations to the side panel, as well as update the source material.

Adds:
- Refresh functionality to the Side Panel
- Refresh functions for Specimen and Digital Media Annotations

Modifies:
- ShowWithAllAnnotations function to become a ShowWithAnnotations function that takes two parameters: optional annotations (default is all annotations), and a target property to keep track of).

### DIFF
--- a/src/api/specimen/GetSpecimenAnnotations.ts
+++ b/src/api/specimen/GetSpecimenAnnotations.ts
@@ -9,9 +9,9 @@ import { SpecimenAnnotations, Annotation, JSONResultArray } from 'global/Types';
 
 
 const GetSpecimenAnnotations = async (handle: string) => {
-    if (handle) {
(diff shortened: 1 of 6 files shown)


TomDijkema:
Again, want to ignore the duplication as I already created a generic component for the Annotation tools, but it still counts the rendering of the component as a duplicated entry in the Specimen and Digital Media pages. This can be solved, but then I would need to make a generic template for the Specimen and Digital Media pages, which I would not be fan off, like to keep them apart.

southeo (approved):
:recycle:

PR #43, opened 2023-08-01
Add image, video, audio and file media types to Specimen Digital Media

TomDijkema:
PR adds templates for images, videos, audio and other media in the shape of files to the Digital Media tab of the Specimen page. Images have been reworked to align vertically (not horizontal in a slider anymore), and include a link to the digital media page. Videos are included and can either can loaded from an external source with the video element, or using an iframe for youtube videos. A custom audio player has been implemented for playing audio. This audio player can be styled freely and should remain the same across different browsers. Lastly, other media is currently housed under the 'other media' category, simply being displayed using a file icon and the name of the media object.

Adds:
- Image type template Specimen Page
- Video type template Specimen Page
- Audio type template Specimen Page
- File type template Specimen Page
- Audio player component

### DIFF
--- a/src/components/general/audioPlayer/AudioPlayer.tsx
+++ b/src/components/general/audioPlayer/AudioPlayer.tsx
@@ -1,0 +1,113 @@
+/* Import Dependencies */
+import { useEffect, useState, useRef, useCallback } from "react";
+import moment from "moment";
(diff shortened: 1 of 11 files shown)


southeo (approved):
:movie_camera: :framed_picture: :musical_keyboard:

PR #47, opened 2023-08-10
Add IIIF viewer

TomDijkema:
PR adds functional IIIF viewer to the Digital Media Page. If a manifest file is provided, the application will try to find the info.json file from the manifest images array and try to open the id url in OpenSeaDragon.

Some institutions are using other implementations of IIIF that do not cohere to the standard set by the IIIF API definition. Possibly notify them on this, because it is a lot of work to implement an exception for every institution that has its own methods.

Adds:
- IIIF viewer on Digital Media Page using OpenSeaDragon


southeo (approved):
:art:

PR #53, opened 2023-09-05
Css clean up

TomDijkema:
PR cleans up a lot of the CSS classes, and instead uses Bootstrap or custom generic classes.


package.json
```diff
@@ -86,5 +86,6 @@
     "@types/openseadragon": "^3.0.6",
     "@types/react-datepicker": "^4.8.0",
     "@types/validator": "^13.7.17"
-  }
+  },
+  "proxy": "https://sandbox.dissco.tech"
```
samleeflang, line 90: Is this needed? And probably switch this to dev.dissco.tech

src/App.css
```diff
@@ -248,7 +221,14 @@ p {
   border-radius: 8px;
 }
 
+.rounded-full {
+  border-radius: 999px;
```
samleeflang, line 225: This looks a bit strange, is 999px correct?

src/components/search/components/searchMenu/SearchBar.tsx
```diff
@@ -36,11 +36,11 @@ const SearchBar = () => {
                 SearchByQuery(form.q);
             }}
         >
-            <Form className={styles.searchBarBlock}>
+            <Form className="pb-3">
                 <Row>
                     <Col className="pe-2">
                         <Field type="text" name="q"
-                            className={`${styles.searchBar} w-100 px-2 py-1`}
+                            className={`${styles.searchBar} rounded-c w-100 px-2 py-1`}
```
samleeflang, line 43: Tom can you explain why in most cases you got rid of the styles object but on other cases we use both the styles object and additional classes.

samleeflang (commented):
Great that we got rid of a lot of css. Got a couple of questions

package.json
TomDijkema: That one slipped through, is a development variable, removed it

src/components/search/components/searchMenu/SearchBar.tsx
TomDijkema: Yes I can! So we still keep some of the custom classes in there as they are necessary, because some CSS attributes are so underused or specific it is not useful to create a custom generic class for it. So, sometimes an element can have both generic class names as a custom one that adds a little bit of specialty.

Another use case is when we use an active class. The active class is used to indicate some element is in an 'active' state and the CSS should change to indicate this activeness. Whilst the class itself is not in the CSS file anymore, the active class is and we need to define the class name in order to get the active effect. That is why we add the custom class on the element, and then only add the active class's effect in the CSS file, like so:

```
.searchBar.active {
    backgroundColor: red;
}
```

src/App.css
samleeflang: :+1:

src/components/search/components/searchMenu/SearchBar.tsx
samleeflang: Ok clear, thanks!

samleeflang (approved):

PR #54, opened 2023-09-05
Download specimen

TomDijkema:
PR adds download option for Specimen and changes several small things for improving the user experience. It changes the Earth System topic discipline to geology. Adds a 'take a tour' button to the header to initiate a tour rather than the compass icon. Changes the downloaded specimen file name to be the PID with the version attached. Lastly, changes annotation motivation to annotation type because the term motivation would be too vague.

### DIFF
--- a/src/components/specimen/components/TitleBar.tsx
+++ b/src/components/specimen/components/TitleBar.tsx
@@ -56,7 +56,7 @@ const TitleBar = (props: Props) => {
         /* Create and click on link to download file */
         const link = document.createElement("a");
         link.href = URL.createObjectURL(jsonFile);
-        link.download = `${specimen.specimenName}.json`;
+        link.download = `${specimen.id.replace('https://hdl.handle.net/', '')}_${specimen.version}.json`;
 
         link.click();
     }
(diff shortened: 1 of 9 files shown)


src/components/specimen/components/TitleBar.tsx
```diff
@@ -41,9 +41,26 @@ const TitleBar = (props: Props) => {
...
+        { value: 'download', label: 'Download as JSON' }
     ];
 
+    /* Function for downloading a Specimen as JSON */
+    const DownloadSpecimen = () => {
+        /* Parse Specimen object to JSON */
+        const jsonSpecimen = JSON.stringify(specimen);
+
+        /* Create JSON file */
+        const jsonFile = new Blob([jsonSpecimen], { type: "text/plain" });
```
samleeflang, line 54: Is text/plain correct? or should it be `"application/json"`

samleeflang (commented):
One comment

src/components/specimen/components/TitleBar.tsx
TomDijkema: Both work for me locally, changed it to 'application/json' to be inline with the API

samleeflang (approved):

PR #57, opened 2023-09-18
Demo doi tooltip

TomDijkema:
PR includes the latest changes to the DOI Tooltip componente.

It also includes a title and short description about the demo and the tooltip component.

Finally, it includes a form field in which the user can manually insert a DOI, on which a tooltip will be created.

### DIFF
--- a/src/components/demo/Demo.tsx
+++ b/src/components/demo/Demo.tsx
@@ -1,4 +1,6 @@
 /* Import Dependencies */
+import { useState } from 'react';
+import { Formik, Form, Field } from 'formik';
 import { Container, Row, Col } from 'react-bootstrap';
 
 /* Import Styles */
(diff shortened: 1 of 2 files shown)


src/components/demo/Demo.tsx
```diff
@@ -11,13 +13,47 @@ import Footer from 'components/general/footer/Footer';
...
+
+                            <p className="text-center">
+                                This demo page demonstrates the DOI Tooltip component offered by DiSSCo.
+                                <br />
+                                It is a component that can be used to cite DOI in text. When clicked on, it displays details on the DOI.
+                                <br />
+                                To implement the component yourself, please refer to this Notebook
+                                <br />
+                                <span className="c-accent">
+                                    <a href="https://colab.research.google.com/drive/1edY_tsIJvtQJsntTHk_xtnbdyuhLw6Br?usp=sharing"
```
samleeflang, line 37: Is this a document on the DiSSCo drive or is it local? Might even be better to drop these documents somewhere in github

TomDijkema: Currently in the DiSSCo Drive, but will also include it here

samleeflang (approved):

PR #60, opened 2023-10-05
Image annotation

TomDijkema:
PR adds Image Annotation to the application.
Annotorious version 3 upon the OpenSeaDragon viewer is chosen as the image annotation tool.
This tool allows users to draw bounding boxes upon an image and add a text value to that area.
These annotations show then up in other peoples view of the image and can also make annotations on it, or add separate annotations on the image for other marks.
Annotations can be edited, as in the bounding box can be moved or reshaped in size, but the value of the annotation can also be updated.
Lastly, visual annotations can be deleted by the owner.

Adds:
- Image Annotation


src/global/Types.ts
```diff
@@ -74,7 +74,7 @@ export interface Specimen {
     type: string
 };
 
-export interface SpecimenDigitalMedia  {
+export interface SpecimenDigitalMedia {
```
samleeflang, line 77: Wondering if we can generate these interface's from the json schema:
https://www.npmjs.com/package/json-schema-to-typescript
That is also what we do in Java, we would then have one schema and generate the types in either java/typescript or python
Not for this PR but maybe create a task for it

src/global/Types.ts
```diff
@@ -149,11 +161,35 @@ export interface AnnotationTemplate {
...
 };
 
+export interface ImageAnnotationTemplate {
+    type: string,
+    motivation: string,
+    body: {
+        type: string,
+        values: string[]
+    },
+    target: {
```
samleeflang, line 176: Looks duplicate

src/components/general/mediaTypes/mediaTypes.module.scss
```diff
@@ -1,3 +1,12 @@
 .organisationLogo {    
     filter: grayscale(100%);
+}
+
+.annotoriousPopUp {
+    width: 400px;
```
samleeflang, line 6: Isn't fixed px width tricky, especially with eye on responsiveness?

src/components/general/mediaTypes/ImagePopup.tsx
```diff
@@ -0,0 +1,174 @@
...
+    RefreshAnnotations: Function,
+    SubmitAnnotation: Function,
+    SetSelectedAnnotation: Function,
+    SetEditAnnotation: Function,
+    RemoveAnnotation: Function
+};
+
+
+const ImagePopup = (props: Props) => {
+    const { selectedAnnotation, editAnnotation, tooltipFieldRef, anno,
```
samleeflang, line 35: What does `anno` mean, maybe improve this variable name

src/components/general/mediaTypes/ImagePopup.tsx
```diff
@@ -0,0 +1,174 @@
...
+                        <Col>
+                            <div className={classEditBlock}>
+                                <Formik
+                                    initialValues={{
+                                        annotationValue: (editAnnotation?.body.length) ?
+                                            editAnnotation.body[0].value[0] : '',
+                                    }}
+                                    enableReinitialize={true}
+                                    validate={ValidateAnnotation}
+                                    onSubmit={async (form) => {
```
samleeflang, line 73: Only authenticated user can add annotations right?

src/components/general/mediaTypes/ImagePopup.tsx
```diff
@@ -0,0 +1,174 @@
...
+                                        </Form>
+                                    )}
+                                </Formik>
+                            </div>
+
+                            <div className={classInfoBlock}>
+                                <Row>
+                                    <Col>
+                                        <p className="fs-4">
+                                            {(selectedAnnotation?.body.length) ?
```
samleeflang, line 119: Do we also show the PID and the other minimal metadata for each annotation in the card?

src/components/general/mediaTypes/ImageViewer.tsx
```diff
@@ -0,0 +1,319 @@
...
+                    const ROI = imageAnnotation.target.selector.hasROI;
+
+                    const x = ROI["ac:xFrac"] * imgWidth;
+                    const y = ROI["ac:yFrac"] * imgHeight;
+                    const w = ROI["ac:widthFrac"] * imgWidth;
+                    const h = ROI["ac:heightFrac"] * imgHeight;
+
+                    visualImageAnnotations.push({
+                        id: imageAnnotation.id,
+                        "@context": 'http://www.w3.org/ns/anno.jsonld',
```
samleeflang, line 191: Not sure about this context field

src/components/general/mediaTypes/ImageViewer.tsx
```diff
@@ -0,0 +1,319 @@
...
+                            type: 'TextualBody',
+                            value: imageAnnotation.body.values,
+                            purpose: 'commenting',
+                            creator: {
+                                id: imageAnnotation.creator
+                            }
+                        }],
+                        target: {
+                            selector: {
+                                conformsTo: 'http://www.w3.org/TR/media-frags/',
```
samleeflang, line 203: I thought we switched to audobon core, or did we go for media-frags?

src/components/general/mediaTypes/ImageViewer.tsx
```diff
@@ -0,0 +1,319 @@
...
+            /* Calculate W3C pixels to TDWG AC position */
+            const coordinates = editAnnotation.target.selector.value.split(',');
+
+            const xFrac = Number(coordinates[0].replace('xywh=pixel:', '')) / image.width;
+            const yFrac = Number(coordinates[1]) / image.height;
+            const widthFrac = Number(coordinates[2]) / image.width;
+            const heightFrac = Number(coordinates[3]) / image.height;
+
+            /* Prepare new Annotation object */
+            const imageAnnotation: ImageAnnotationTemplate = {
```
samleeflang, line 229: What is the difference with the above one?

samleeflang (commented):
Couple of comments / questions

package.json
```diff
@@ -8,6 +8,7 @@
     ]
   },
   "dependencies": {
+    "@annotorious/react": "^3.0.0-pre-alpha-47",
```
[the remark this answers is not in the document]
TomDijkema: Updated the ADR: https://docs.google.com/document/d/1XfJgB-_DmArevxZ0DlTd6Cn3VK5BL5igQoQ9O5W-cZs/edit#heading=h.f97858opk6yd

src/components/digitalMedia/DigitalMedia.tsx
```diff
@@ -248,6 +267,26 @@ const DigitalMedia = () => {
                                                             />
                                                         </Col>
                                                         <Col />
+                                                        {(digitalMedia.type === '2DImageObject' && KeycloakService.IsLoggedIn()) &&
```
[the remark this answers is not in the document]
TomDijkema: Yes, I expected that, there is a lot that needs to be changed when the new data model drops

src/components/digitalMedia/components/annotate/test.css
```diff
@@ -0,0 +1,23 @@
+.helloworld-widget {
```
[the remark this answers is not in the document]
TomDijkema: Good catch, removed this file

src/global/Types.ts
TomDijkema: Interesting thought. I would have to download the JSON schemas in order to use this I think.
https://naturalis.atlassian.net/browse/DD-736?atlOrigin=eyJpIjoiZDNiZThlMjQwZGQ4NGNlMjg3MDQxOTYxOTFmNDEyYmUiLCJwIjoiaiJ9

src/global/Types.ts
TomDijkema: Yeah kinda, I will replace this in the upcoming PR for the implementation of the Annotation model

src/components/general/mediaTypes/mediaTypes.module.scss
TomDijkema: Yes in principle, this element is positioned absolute, which means it will hover over everything but in a certain position.
Setting a certain width is needed for absolute elements, percentages do work, but then it needs to be position relative to the parent, in this case the annotorious component.
In version 2 this pop-up was automatically appointed to a width of 400px, that is why I took over this value.
We can make it responsive, but I doubt it would be better for this specific case.

src/components/general/mediaTypes/ImagePopup.tsx
TomDijkema: It basically holds the Annotorious object, used the variable as it is also called 'anno' in the author's code.
Changed it to 'annotorious'

src/components/general/mediaTypes/ImagePopup.tsx
TomDijkema: Yes Sir

src/components/general/mediaTypes/ImagePopup.tsx
TomDijkema: At the moment only the value, I can add the PID, would we want to add all the metadata like in the side panel?

src/components/general/mediaTypes/ImageViewer.tsx
TomDijkema: This is to feed Annotorious, for our own annotations we use: 
`conformsTo: 'https://ac.tdwg.org/termlist/#711-region-of-interest-vocabulary'`

src/components/general/mediaTypes/ImageViewer.tsx
TomDijkema: Also to feed Annotorious, we might be able to use another context, I took this over from the author's code

src/components/general/mediaTypes/ImageViewer.tsx
TomDijkema: The one above is for feeding Annotorious, they use the w3 format and use a kinda different structured object.
The one below is for saving the annotation in our system, calculate the TDWG ac standard from the w3 standard (pixels to position).

src/components/general/mediaTypes/mediaTypes.module.scss
samleeflang: Ok sure, let's work with this and change it if we hit issues

src/components/general/mediaTypes/ImagePopup.tsx
samleeflang: Yes that would be great! Maybe in a separate story so we can close this PR

src/components/general/mediaTypes/ImageViewer.tsx
samleeflang: ah right, thanks for explaining

src/components/general/mediaTypes/ImageViewer.tsx
samleeflang: Yes thanks I forgot that annototious also uses the w3 format

samleeflang (approved):
Approved, showing the annotation metadata in the annotoroius viewer will be part of a new PR.

PR #62, opened 2023-10-16
Media annotation template&bug fixes

TomDijkema:
PR updates the Annotorious package to resolve the broken image annotation and resolves a bug with the multi select filter chosen options to be removed when not part of a search query upon that filter.


src/components/search/components/searchMenu/filters/MultiSelectFilter.tsx
```diff
@@ -81,11 +81,13 @@ const MultiSelectFilter = (props: Props) => {
 
     /* OnChange of Search Query: refresh aggregations by query */
     useEffect(() => {
-        const copyAggregations = { ...aggregatons };
+        const copyAggregations = { ...aggregations };
 
         /* Function to Refresh Aggregations */
         const RefreshAggregations = () => {
             GetSpecimenAggregations(searchFilters).then((aggregations) => {
+                console.log(aggregations);
```
samleeflang, line 89: Couple of log statements. Do want to keep those or were they only for testing?

TomDijkema: Good catch

samleeflang (approved):

PR #64, opened 2023-10-19
Scale Specimen ID Card and Text / change the Digital Specimen and Med…

TomDijkema:
PR scales the Specimen ID Card to fit its properties, the height is divided equally. Fixed the banner image not showing up, and also implemented a scalar for the general text in the application to make the interface appear the same, as well on smaller as bigger screens.

Commit also changes the digital specimen and media types. They now include the different levels of nesting as present in the API. This does require  an extra level of assigning or requesting properties, but should be better as it is inline with the source, namely: the API

### DIFF
--- a/src/components/specimen/Specimen.tsx
+++ b/src/components/specimen/Specimen.tsx
@@ -219,8 +219,7 @@ const Specimen = () => {
                         />
 
                         <div className="flex-grow-1 overflow-hidden">
-                            {(specimen.digitalSpecimen['ods:id'] &&
-                                specimen.digitalSpecimen['ods:id'].replace('https://doi.org/', '') === `${params['prefix']}/${params['suffix']}`) &&
+                            {(specimen.digitalSpecimen['ods:id'] && specimen.digitalSpecimen['ods:id'].replace('https://doi.org/', '') === `${params['prefix']}/${params['suffix']}`) &&
                                 <Container fluid className="h-100 pt-5">
                                     <Row className="h-100">
                                         <Col className={`${classSpecimenContent} h-100 transition`}>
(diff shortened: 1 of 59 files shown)


src/components/specimen/Specimen.tsx
```diff
@@ -203,51 +204,56 @@ const Specimen = () => {
...
+                        <SpecimenSteps SetSelectedTab={(tabIndex: number) => setSelectedTab(tabIndex)} />
+                        <AnnotateSteps ShowWithAnnotations={(annotations?: SpecimenAnnotations, property?: string) => ShowWithAnnotations(annotations, property)} />
+                        <MASSteps automatedAnnotationsToggle={automatedAnnotationsToggle}
+                            SetAutomatedAnnotationsToggle={(toggle: boolean) => setAutomatedAnnotationToggle(toggle)}
+                            ShowWithAnnotations={() => ShowWithAnnotations()}
+                        />
+
+                        <div className="flex-grow-1 overflow-hidden">
+                            {(specimen.digitalSpecimen['ods:id'] &&
+                                specimen.digitalSpecimen['ods:id'].replace('https://doi.org/', '') === `${params['prefix']}/${params['suffix']}`) &&
```
samleeflang, line 223: Looks like a white line that doesn't need to be there

samleeflang (commented):
Not much got a question and a small format thing

src/components/specimen/Specimen.tsx
TomDijkema: Did this on purpose because the line got too long in the editor, but reverted this to one line

samleeflang (approved):

PR #66, opened 2023-10-26
Implement new annotation model

TomDijkema:
PR that implements the new Annotation data model.
Implements the new data model by generating the Annotation type based upon the JSON schema.
Adapts the code to make use of the new model, with all the different prefixes included.

PR also reworks the Prompt Message component to make use of the Bootstrap Toast component.


src/components/general/nomenclatural/Class.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+
+const Class = (props: Props) => {
+    const { classProp } = props;
+
+    /* Base variables */
+    let classElement: ReactElement;
+
+    /* Format Class according to Nomenclatural Rules */
+    classElement = <span> {classProp ?? 'undefined'} </span>
```
samleeflang, line 18: Don't think this should be `undefined` but be left empty. If it stays undefined we should keep it in line with the other Undefined which I believe have a capital

src/components/specimen/steps/AnnotateSteps.tsx
```diff
@@ -42,21 +42,31 @@ const AnnotateSteps = (props: Props) => {
...
     const dummyAnnotation: Annotation = {
-        id: '20.5000.1025/dummy',
-        version: 1,
-        type: 'Annotation',
-        motivation: 'commenting',
-        target: {
-            id: specimen.digitalSpecimen['ods:id'],   
-            type: 'digitalSpecimen',
-            indvProp: 'ods:specimenName'
+        "ods:id": '20.5000.1025/dummy',
```
samleeflang, line 45: maybe make the prefix TEST or something else

src/sources/annotationMotivations.json
```diff
@@ -1,37 +1,30 @@
 {
-    "adding": {
-        "key": "adding",
+    "oa:adding": {
```
samleeflang, line 2: I think this one is `ods:adding`

src/components/specimen/Specimen.tsx
```diff
@@ -122,11 +122,11 @@ const Specimen = () => {
         const copySpecimenAnnotations = { ...specimenAnnotations };
 
         /* Check if array for target property exists */
-        if (annotation.target.indvProp in specimenAnnotations) {
+        if (annotation['oa:target']['oa:selector']?.['ods:field'] as string in specimenAnnotations) {
```
samleeflang, line 125: Why is this necessary?

samleeflang (commented):
Great work! Main thing I am missing is the ClassSelector stuff but we dan first do the new model and then get it in there. Do you have user stories for implementing ClassSelector.

src/api/digitalMedia/GetDigitalMediaAnnotations.ts
```diff
@@ -26,19 +24,17 @@ const GetDigitalMediaAnnotations = async (handle: string) => {
...
 
             /* Refactor Annotations object */
             annotations.forEach((annotation) => {
-                if (annotation.target.selector?.hasROI) {
+                if (annotation['oa:target']['oa:selector']?.['ac:hasROI']) {
                     digitalMediaAnnotations.visual.push(annotation);
-                } else if (digitalMediaAnnotations[annotation.target.indvProp]) {
-                    digitalMediaAnnotations[annotation.target.indvProp].push(annotation);
+                } else if (digitalMediaAnnotations[annotation['oa:target']['oa:selector']?.['ods:field'] as string]) {
+                    digitalMediaAnnotations[annotation['oa:target']['oa:selector']?.['ods:field'] as string].push(annotation);
```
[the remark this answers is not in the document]
TomDijkema: Yes, I have not started with the ClassSelector topic. First implement the data model things and then continue with these kind of advancements

src/components/general/nomenclatural/Class.tsx
TomDijkema: All 'Undefined' string are now empty strings, also this one with the lowercase 'u'

src/components/specimen/steps/AnnotateSteps.tsx
TomDijkema: Changed to TEST

src/sources/annotationMotivations.json
TomDijkema: You are right, changed to 'ods:adding'

src/components/specimen/Specimen.tsx
TomDijkema: Well this is a fun one. It gets complicated because I want to access a one of optional property inside of an already optional property. TypeScript gets really nervous when I do not include the optional (?) and the type definition of the optional one of property. Especially the one of is triggering TypeScript as it can be a string, but also unknown (not there at all). That's why I need to declare it to be a string, assuming it to be there and interpreting it as a string.

samleeflang: Ok thanks, sounds like fun

samleeflang (approved):
:1st_place_medal:

PR #67, opened 2023-10-27
Data model specimen occurrences tab

TomDijkema:
PR adds Properties Table for displaying all properties assigned to the different data groups within the new Specimen data model.
Select a property from this table to open the Annotation Side Panel and make annotations.
The table is first applied in the Occurrences tab, now displaying all occurrences per card, distinguishing the different sub groups as separate tables.
There are:
- occurrences main level properties
- location
- georeference
- geological context
- assertions


src/components/specimen/components/contentBlocks/Occurrences.tsx
```diff
@@ -0,0 +1,117 @@
...
+    let occurrences: {
+        properties: Dict,
+        location?: Dict,
+        georeferece?: Dict,
+        geologicalContect?: Dict,
+        assertions?: Dict
+    }[] = [];
+
+    /* Craft occurrences object to iterate over */
+    specimen.digitalSpecimen.occurrences?.forEach((occurrence, index) => {
```
samleeflang, line 37: Why is this needed?

samleeflang (commented):
Just one question out of curiosity.

samleeflang (approved):

PR #68, opened 2023-10-27
Data model specimen identifications tab

TomDijkema:
PR adds content to the Identifications tab. The Properties Table is used to to display the properties. As an identification can have multiple taxon identifications, these are separated and shown in separate tables.

Adds:
- Identifications tab with content (identifications)

### DIFF
--- a/src/components/specimen/components/contentBlocks/Identifications.tsx
+++ b/src/components/specimen/components/contentBlocks/Identifications.tsx
@@ -81,7 +81,7 @@ const Identifications = (props: Props) => {
                                             <div key={identificationKey} className="mt-3">
                                                 <PropertiesTable
                                                     title={identificationKey}
-                                                    properties={identification[identificationKey as keyof typeof identification] as Dict}
+                                                    properties={identification[identificationKey]}
                                                     ShowWithAnnotations={(property: string) => ShowWithAnnotations(property)}
                                                 />
                                             </div>
--- a/src/components/specimen/components/contentBlocks/Identifications.tsx
+++ b/src/components/specimen/components/contentBlocks/Identifications.tsx
@@ -1,0 +1,100 @@
+/* Import Dependencies */
+import { isEmpty, cloneDeep } from 'lodash';
+import classNames from 'classnames';
+import { Card, Row, Col } from 'react-bootstrap';
+
+/* Import Store */
+import { useAppSelector } from 'app/hooks';
+import { getSpecimen } from 'redux/specimen/SpecimenSlice';
+
+/* Import Types */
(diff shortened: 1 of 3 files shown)


src/components/specimen/components/contentBlocks/Identifications.tsx
```diff
@@ -0,0 +1,100 @@
...
+        const copyIdentification = cloneDeep(identification);
+
+        /* Push new craft occurrence to array */
+        identifications.push({ properties: {} });
+
+        /* Check for Taxon Identifications */
+        if (!isEmpty(identification.taxonIdentifications)) {
+            identification.taxonIdentifications?.forEach((taxonIdentification, index) => {
+                /* Add Taxon Identification to craft identification */
+                identifications[index][`taxon Identification #${++index}`] = taxonIdentification;
```
samleeflang, line 44: Maybe also capitalize `taxon`

src/components/specimen/components/contentBlocks/Identifications.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+    /* Base variables */
+    const specimen = useAppSelector(getSpecimen);
+    let identifications: {
+        properties: Dict,
+        [taxonIdentification: string]: Dict
+    }[] = [];
+
+    /* Craft identifications object to iterate over */
+    specimen.digitalSpecimen['dwc:identification']?.forEach((identification, index) => {
```
samleeflang, line 34: So you basically flatten the object so that it fits easier into the table. Wonder if we can have collapsable sections in the table to separate the different identifications. Maybe something to discuss/brainstorm

src/components/specimen/components/contentBlocks/Identifications.tsx
```diff
@@ -0,0 +1,100 @@
...
+                    'mt-3': index > 0
+                });
+
+                return (
+                    <Card key={key} className={CardClass}>
+                        <Card.Body>
+                            <Row>
+                                <Col>
+                                    <p className="fs-2 fw-lightBold">
+                                        {identification.properties['dwc:identificationVerificationStatus'] ?
```
samleeflang, line 70: Oh yes nice it already splits them and adds a Header :+1:

src/components/specimen/components/contentBlocks/Identifications.tsx
TomDijkema: Yeah exactly, also tried to explain this in the PR of the occurrences tab. Yes that is a good idea to make them collapsable, will add that task to the board

samleeflang (approved):

PR #69, opened 2023-11-08
Data model specimen entity relations tab

TomDijkema:
PR adds Entity Relationships tab containing two different views: a table view for displaying all the data in an orderly manner, and a graph view for visually indicating the relations to the specimen. A relation can be selected for more information. This still needs to be filled. The styling also needs to be reviewed.

Adds:
- Entity Relationships tab
- Table view for entity relationships
- Graph view for entity relationships


samleeflang (approved):
Looks good, can't wait to see it in action. Might be best to first merge the other two.

PR #71, opened 2023-11-13
Visual annotations meta data

TomDijkema:
PR adds additional metadata to a Visual Annotation. The metadata, and appearance, is the same as the annotations displayed in the Side Panel. Also adds global definitions in the .env file for doi.org and handle.net as these strings are repeated a lot in the overall code. Now, the variable takes care of it. Also, if in any case the id's will not be DOIs and hosted somewhere else, the url can be changed rapidly.

Adds:
- Additional metadata to Visual Annotation
- Global definitions for doi.org and handle.net variables


src/components/general/mediaTypes/ImagePopup.tsx
```diff
@@ -112,39 +132,46 @@ const ImagePopup = (props: Props) => {
...
-                                                <FontAwesomeIcon icon={faTrashCan}
-                                                    className="c-pointer"
-                                                    onClick={() => RemoveAnnotation()}
-                                                />
-                                            </Col>
-                                        </>
-                                    }
-                                </Row>
-                            </div>
+                            {annotation &&
```
samleeflang, line 135: It wasn't possible to reuse the same component?

src/components/general/mediaTypes/ImagePopup.tsx
```diff
@@ -55,6 +73,8 @@ const ImagePopup = (props: Props) => {
         'd-none': editAnnotation
     });
 
+    console.log(annotation);
```
samleeflang, line 76: Can this be removed?

samleeflang (commented):
:t-rex:  Small comment and a question.

src/components/general/mediaTypes/ImagePopup.tsx
TomDijkema: Yeah let's do that, I will pick this up after the current stuff then

src/components/general/mediaTypes/ImagePopup.tsx
TomDijkema: Removed

samleeflang (approved):

PR #72, opened 2023-11-28
Annotation class selector

TomDijkema:
A Pull Request that completely changes up the Annotation game within DiSSCover.
The changes makes the annotation process much more rich and completely dependent on the data model that is fed to the system; any changes in the data model like additions of properties or classes will be taken over without any problem.
The biggest change surrounds the addition of the class selector for annotations.

Adding the class selector meant a lot of things had to shift around to make the nesting, required for accessing the properties of a class, sub class or array of sub classes possible.
A crucial distinction now made in the Annotate Target state variable is the targetField and targetClass.

A field always points to a single property of an object (class), and is easy annotatable, only needing a motivation, annotation value and possible additional fields based upon the chosen motivation.

A class points to an object, containing multiple properties, sub classes or even arrays of sub classes.
Its far more complicated than the field, but the two target types all share the same parent (state) variables and template files, this enlarges the files themselves, but keeps all the relevant code together.
Since there can be multiple of the same classes within a Digital Specimen or Digital Media, an user will always annotate a certain instance of a class, say the third occurrence of a specimen, to catch the correct occurrence, the whole path to this sub class is remembered as the 'target'; JSON path is used for the syntax.
This path also includes indexes for arrays to be able to indicate the right number of an array instance.

Annotating a field or class will now first search for existing instances within the Digital Specimen or Digital Media and return a list of these.
The user can decide to add a completely new instance of this field or class, or annotate on of the existing ones.
Annotating a class with the adding or editing motivations will generate a complete list of all the properties within the class which can be given values; opening the gateway for users to add completely new occurrences for example to the Digital Specimen.

Even though the annotation process should be relatively flexible from now on, it still requires review on how to act in certain scenarios.
Updates will keep coming to this topic.

### DIFF
--- a/src/app/Utilities.ts
+++ b/src/app/Utilities.ts
@@ -1,3 +1,7 @@
+/* Import Types */
+import { Dict } from "./Types";
+
+
 /* Function to capitalize the first character of a string */
 const Capitalize = (string: string) => {
     if (string) {
@@ -56,7 +60,7 @@ const GetFilters = (searchParams: URLSearchParams) => {
 const RandomString = () => {
     let result = '';
 
-    const characters ='ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+    const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
     const charactersLength = characters.length;
 
     for (let i = 0; i < 10; i++) {
@@ -66,10 +70,34 @@ const RandomString = () => {
     return result;
 }
 
+/* Function for returning all key value pairs from an object, including nested ones */
+const ReturnPropertiesFromNestedObject = (propertiesObject: Dict) => {
+    const propertiesArray: { key: string, value: string | number | boolean }[] = [];
+
+    const PushToPropertiesArray = (propertyPair: { key: string, value: string | number | boolean }) => {
+        propertiesArray.push(propertyPair);
+    }
+
+    const LoopOverProperties = (propertiesObject: Dict) => {
+        for (const property in propertiesObject) {
+            if (typeof propertiesObject[property] !== 'object') {
+                PushToPropertiesArray({ key: property, value: propertiesObject[property] });
+            } else {
+                LoopOverProperties(propertiesObject[property]);
+            }
+        }
+    }
+
+    LoopOverProperties(propertiesObject);
+
+    return propertiesArray;
+}
+
 export {
     Capitalize,
     CheckProperty,
     DetectMobile,
     GetFilters,
-    RandomString
+    RandomString,
+    ReturnPropertiesFromNestedObject
 };
--- a/src/app/utilities/AnnotationUtilities.ts
+++ b/src/app/utilities/AnnotationUtilities.ts
@@ -15,9 +15,13 @@ import ChronometricAge from 'sources/dataModel/chronometric_age.json';
 import Location from 'sources/dataModel/location.json';
 
 
-const ConstructList = (schema: string, properties: Dict, classValue: string, PushToPropertiesList: Function, PushToClassesList: Function, subSchema?: string) => {
+/* Recursive function for constructing a list of annotatable properties and classes */
+const ConstructList = (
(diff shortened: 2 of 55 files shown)


src/app/Utilities.ts
```diff
@@ -66,10 +70,34 @@ const RandomString = () => {
...
 
+/* Function for returning all key value pairs from an object, including nested ones */
+const ReturnPropertiesFromNestedObject = (propertiesObject: Dict) => {
+    const propertiesArray: { key: string, value: string | number | boolean }[] = [];
+
+    const PushToPropertiesArray = (propertyPair: { key: string, value: string | number | boolean }) => {
+        propertiesArray.push(propertyPair);
+    }
+
+    const LoopOverProperties = (propertiesObject: Dict) => {
```
samleeflang, line 81: Recursion looks fine :+1: What we could do is write a unit test for these type of functions just to make sure they work and keep working

src/app/utilities/AnnotationUtilities.ts
```diff
@@ -0,0 +1,228 @@
...
+import Assertion from 'sources/dataModel/assertions.json';
+import Occurrence from 'sources/dataModel/occurrences.json';
+import EntityRelationship from 'sources/dataModel/entity_relationships.json';
+import Citation from 'sources/dataModel/citations.json';
+import Identifier from 'sources/dataModel/identifiers.json';
+import ChronometricAge from 'sources/dataModel/chronometric_age.json';
+import Location from 'sources/dataModel/location.json';
+
+
+/* Function for pushing to the annotatable properties list, or recall Construct List */
```
samleeflang, line 18: Could you explain this one to me, trying to wrap my mind around it but having some difficulty

src/components/annotate/sidePanel/SidePanel.tsx
```diff
@@ -164,30 +164,43 @@ const SidePanel = (props: Props) => {
...
+                                                    {`${annotateTarget.targetProperty.name ?
+                                                        annotateTarget.targetProperty.name
+                                                        : editAnnotation['oa:target']['oa:selector']?.['ods:field'] as string}`
+                                                    }
+                                                </>
+                                            }
+                                        </p>
+                                    </Col>
+                                </Row>
+                                {/* {(annotateTarget.targetProperty.name || editAnnotation['oa:target']['oa:selector']?.['ods:field'] as string) &&
```
samleeflang, line 196: Is it ok that this is commented?

src/app/Utilities.ts
TomDijkema: Will keep this in mind, might be good as its a pure functional function

src/app/utilities/AnnotationUtilities.ts
TomDijkema: Ah yes, so this file has multiple purposes and was created to relieve the template files of all these additional functions.
Primary function of this file is to create the list of annotatable properties and classes, the dropdown lists you will see when opening the Side Panel.
Field properties are grouped by and belong to a class, so simply looping over all the data would not fit the needs, also properties might change by name or classes by structure, hence it start blank with the base files from the data model.

- Main function for creating the lists is ConstructTargetPropertiesLists, which bases the base model (Digital Specimen or Digital Media) and the lists with according function to push to these lists. Then goes on to call ConstructList to actually construct the list from the base model (yes good function names, I know).

- ConstructList goes on to create the full field properties and classes lists. It loops over the different levels in the data model (classes), then calls on PushToClassesList, which checks if we are interacting with a property or a class, based on the type. If it is a string, integer or boolean, it gets pushed to the propertiesList at the right level (called schema and subSchema in code). If it encounters another object, it will go on to create a new level and repeat the process by calling this ConstructList.

- ExtractFromSchema: a function that is used multiple times and also exported from this file for usage elsewhere. It takes a parameter which should point to one of the existing schemas of the data model. If it recognizes one, it will request all the properties from that schema and return them.

Overall it is a lot of code for two simple looking dropdowns, but it also builds the foundation for handling annotations, defining the JSON routes by extracting properties and classes from the different levels of the data model.

src/components/annotate/sidePanel/SidePanel.tsx
TomDijkema: Removed

PR #73, opened 2023-12-05
Overall polishment and Relation Graph using Cytoscape

TomDijkema:
PR adds overall polishment to the application.
Replaced the old relational graph library with Cytoscape, which should be a bit easier to work with and add images and colors. Graphs now also show annotations as potential linkages.

More polishing will be coming, this PR is to save in between.

### DIFF
--- a/src/components/profile/components/user/Passport.tsx
+++ b/src/components/profile/components/user/Passport.tsx
@@ -151,7 +151,7 @@ const Passport = () => {
             {!editMode &&
                 <Row>
                     <Col lg={{ span: 8, offset: 2 }}>
-                        <OrganisationLogoImage organisationId={userProfile.organisation?.replace('https://ror.org/', '')} />zz
+                        <OrganisationLogoImage organisationId={userProfile.organisation?.replace('https://ror.org/', '')} />
                     </Col>
                 </Row>
             }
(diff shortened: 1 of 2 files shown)


src/components/profile/components/user/Passport.tsx
```diff
@@ -151,7 +151,7 @@ const Passport = () => {
             {!editMode &&
                 <Row>
                     <Col lg={{ span: 8, offset: 2 }}>
-                        <OrganisationLogoImage organisationId={userProfile.organisation?.replace('https://ror.org/', '')} />
+                        <OrganisationLogoImage organisationId={userProfile.organisation?.replace('https://ror.org/', '')} />zz
```
samleeflang, line 154: There is `zz` here is that intentional?

src/components/search/components/IDCard/MapMediaExt.tsx
```diff
@@ -64,47 +69,53 @@ const MapMediaExt = (props: Props) => {
...
+                                            accessUri={mediaItem.digitalEntity['ac:accessUri']}
+                                            format={mediaItem.digitalEntity['dcterms:format']}
+                                        />
+                                    </div>
                                 );
                             })}
-                        </div>
-                    </Col>
+                        </Col>
+                    {/* </Col> */}
```
samleeflang, line 118: This Col can be removed?

src/components/profile/components/user/Passport.tsx
TomDijkema: Good catch! Removed

samleeflang (approved):
:sauropod:

PR #77, opened 2023-12-19
Machine Job Record overview

TomDijkema:
PR adds support for the Machine Job Record.
Machine Job Records created for Specimen or Digital Media can be viewed inside the MAS modal.
The overview should help indicate the current state of the MAS that are running in the background, or have been completed.
After scheduling a MAS, the user will be redirected to the overview.

An overview of the Machine Job Records the user set in motion can be found on the Profile page.

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -336,6 +336,10 @@ p {
   font-weight: 500;
 }
 
+.secondaryButton.disabled {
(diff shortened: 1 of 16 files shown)


southeo (approved):
:cactus:

PR #78, opened 2023-12-20
Annotation improvements

TomDijkema:
PR should improve the overall Annotation logic.
Most of the refinements are related to constructing the correct target property paths.
This is a difficult process because of the different levels class and field annotations work on.
These changes make sure to interconnect the two, making annotations on an existing field, will now use the parent class of that field to determine the formatted path using a new utility function.
The most important part of this function is to add the index when necessary, like for example when you are annotating the country field of a location, the user must choose an existing location class, which is nested inside an occurrence class.
Choosing a specific instance will grab the parent classes, check the indexes, and make sure they are represented in the path, resulting with for example: occurrences[0].location.country.

The code that makes up the annotation process has become big and a bit messy, it might be good to clean this up some time in the future. At least the functionality should work.

### DIFF
--- a/src/api/specimen/GetSpecimenFull.ts
+++ b/src/api/specimen/GetSpecimenFull.ts
@@ -44,13 +44,15 @@ const GetSpecimenFull = async (handle: string) => {
             /* Refactor Annotations for Specimen Page */
             if (annotations) {
                 annotations.forEach((annotation) => {
-                    const annotationIndicator: string = annotation['oa:target']['oa:selector']?.['ods:field'] as string
-                        ?? annotation['oa:target']['oa:selector']?.['oa:class'] as string;
+                    let annotationIndicator: string = (annotation['oa:target']['oa:selector']?.['ods:field'] as string)
+                        ?? (annotation['oa:target']['oa:selector']?.['oa:class'] as string);
 
-                    if (annotationIndicator.replace('$.', '') as string in specimenAnnotations) {
-                        specimenAnnotations[annotationIndicator.replace('$.', '')].push(annotation);
+                    annotationIndicator.replace('$.', '');
+
+                    if (annotationIndicator in specimenAnnotations) {
+                        specimenAnnotations[annotationIndicator].push(annotation);
                     } else {
-                        specimenAnnotations[annotationIndicator.replace('$.', '')] = [annotation];
+                        specimenAnnotations[annotationIndicator] = [annotation];
                     }
                 });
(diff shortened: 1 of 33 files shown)


src/app/utilities/AnnotationUtilities.ts
```diff
@@ -221,6 +221,58 @@ const SearchNestedLevels = (level: Dict | Dict[], nestingLevels: string[], targe
...
+    /* Function to concat level to field, add dot if index is greater than zero */
+    const ConcatToField = (level: string, index: number) => {
+        if (index > 0) {
+            formattedPath = formattedPath.concat(`.${level}`);
+        } else {
+            formattedPath = formattedPath.concat(level);
+        }
+    }
+
+    const levels = path.replace(`${targetType}.`, '').split('.');
```
southeo, line 256: I think you're replacing/removing a `.` then splitting on that character. Is that intentional?

src/api/specimen/GetSpecimenFull.ts
```diff
@@ -47,7 +47,7 @@ const GetSpecimenFull = async (handle: string) => {
                     const annotationIndicator: string = annotation['oa:target']['oa:selector']?.['ods:field'] as string
                         ?? annotation['oa:target']['oa:selector']?.['oa:class'] as string;
 
-                    if (specimenAnnotations[annotation['oa:target']['oa:selector']?.[annotationIndicator] as string]) {
+                    if (annotationIndicator.replace('$.', '') as string in specimenAnnotations) {
```
southeo, line 50: Why do you have the line `annotationIndicator.replace('$.', '')` in the condition as well as in the `else` block (line 53)

src/app/utilities/AnnotationUtilities.ts
TomDijkema: Ah yes, I can understand the confusion, this is to make sure that the top level component of a JSON path is removed.
Top level meaning 'DigitalSpecimen' or 'DigitalEntity' for example.
The targetType variable represents that top level, so I am clearing 'DigitalSpecimen' plus a '.' (dot) from the path, leaving the dot out would result in something like this: '.occurrences.' with the dot still in place at the start

src/api/specimen/GetSpecimenFull.ts
TomDijkema: You have a point, added the replace statement before the conditional to make it globally

PR #80, opened 2024-01-02
Review scaling

TomDijkema:
PR reviews the scaling of the application.
The Digital Media has been adjusted to scale better, also the specimen type blocks on the homepage, as well as the version select and actions dropdowns.

As a side note: the triggering of the side panel on the Digital Media page has been adjusted to mimic the behavior of the Specimen page, implementing the new method.


southeo (approved):
:balance_scale: 
(get it...scales?)

PR #82, opened 2024-01-05
Overall improvements

TomDijkema:
PR adds overall improvements based upon observed issues and feedback.

Includes:
- Natural Origin as block on Home
- Change 'specimens' to 'all specimens' in nav
- Disable languages as there is no full support yet
- Use bellis perennis as the default specimen for search placeholders and in tours
- Fix most of the tours based upon the new content (annotation tour has to be looked at in the future)
- Change Footer content
- Add Terms of use page
- Add new columns to search results table

More PRs with overall improvements will come


samleeflang (approved):
Didn't see anything strange, thanks for the fixes!

PR #83, opened 2024-01-11
Migrate to Tan Stack table - Custom Data Table

TomDijkema:
Proud to announce the new and improved Data Table for global usage within DiSSCover.
The new data table component is built by using the Tan Stack table library, which is a headless library.
Headless means that it does not provide a full component to be used, but rather an API that can be requested and used for passing functionality onto a plain HTML table.

The custom Data Table component has several options built in at the moment:
- Functionality for filtering per column
- Sticky header when scrolling down
- Row selection
- Pinned columns
- Dynamic width of table columns based upon REM value

The columns and data to pass to the data table should be defined loosely from the component itself.
That is why table configuration files have been introduced to take care of the column configuration.
Most of this code is repeatable and takes a lot of space, these files should help prevent code from becoming messier.

New features to the Data Table component may be introduced in the future by demand.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@fortawesome/react-fontawesome": "^0.1.19",
         "@react-keycloak/web": "^3.4.0",
         "@reduxjs/toolkit": "^1.9.5",
+        "@tanstack/match-sorter-utils": "^8.11.3",
         "@tanstack/react-table": "^8.11.3",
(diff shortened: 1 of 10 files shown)


TomDijkema:
> ⚡ Ok, I don't know if I get it all but didn't see anything strange. Looking forward to seeing it in action

I can maybe explain it on Thursday, when you see the logic, it is pretty straightforward!

PR #84, opened 2024-01-17
Overall improvements + Annotorious rewrites

TomDijkema:
PR adds overall improvements:
- Utility function for adding spaces before capitals
- Apply new utility function to filter tags
- Make text in compare box static
- Add DOI to Specimen ID Card header on Search page
- Fix annotation path generation for indexes from the annotation form and overview tables

PR also revisits the code surrounding the Annotorious implementation.
Removes the state variables holding annotation information and instead directly uses the Annotorious API for retrieving information about the canvas and its annotations.


samleeflang (commented):
Changes look good! Couple questions/comments.

src/app/Utilities.ts
```diff
@@ -11,6 +11,21 @@ const Capitalize = (string: string) => {
     }
 }
 
+/* Function to replace capitals in a string with spaces */
+const AddSpaceBeforeCaptials = (string: string) => {
```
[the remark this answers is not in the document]
TomDijkema: Yes, actually I always use capitals for functions and start variables with lower characters to keep them separated easily, this only does not apply to state setting functions because they always start with 'set' or 'get'.
So I guess functions are in PascalCase. Always weird it is called CamelCase, because camels have two humps and you would assume the first also to be in capitals, but I think 'Dromedaris Case' does not sound that good :camel:

src/components/specimen/steps/AnnotateSteps.tsx
```diff
@@ -273,6 +273,12 @@ const AnnotateSteps = (props: Props) => {
                     clearInterval(targetInterval);
                     UpdateAnnotationView(dummyAnnotation, true);
                 }}
+                options={{
```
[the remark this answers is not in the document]
TomDijkema: Made a central config file for the options part, we could later build further on this by including more options or even the JSON data in there or something

src/components/general/mediaTypes/ImagePopup.tsx
```diff
@@ -26,27 +26,30 @@ import { faPencil, faTrashCan } from '@fortawesome/free-solid-svg-icons';
...
     const user = useAppSelector(getUser);
-    let annotation: AnnotationType | undefined;
+    const digitalMediaAnnotations = useAppSelector(getDigitalMediaAnnotations);
+    const [editMode, setEditMode] = useState<boolean>(false);
+    const annotoriousAnnotation = annotorious.getSelected()[0];
+    const selectedAnnotation: Annotation | undefined = digitalMediaAnnotations.visual.find((annotation) => annotation['ods:id'] === annotoriousAnnotation.id);
+
+    /* OnChange of Selected Annotation in Annotorious: check if it is a template for a new Annotation, if so, set Edit Mode to true */
+    useEffect(() => {
+        if (annotoriousAnnotation && !annotoriousAnnotation.id.includes('20.5000.1025') && !annotoriousAnnotation.id.includes('TEST')) {
```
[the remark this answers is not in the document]
TomDijkema: Replaced this with a check for the '/' character. This character does not seem to appear in the ids generated by Annotorious, but is always present in ours, separating the prefix from the suffix, so we are able to check by this.

src/components/annotate/sidePanel/Annotation.tsx
```diff
@@ -58,11 +58,11 @@ const Annotation = (props: Props) => {
...
 
     /* Transform array of values to displayable information */
-    if (annotation['oa:target']['oa:selector']?.['oa:class']) {
+    try {
         const annotationValueObject = JSON.parse(annotation['oa:body']['oa:value'].join(', '));
 
         annotationValue = annotationValueObject[(annotation['oa:target']['oa:selector']?.['oa:class'] as string).replace('$.', '') as keyof typeof annotationValueObject];
-    } else {
+    } catch {
         annotationValue = annotation['oa:body']['oa:value'].join(', ');
```
samleeflang, line 66: This will only happen when the JSON is malformed right? Maybe also log the exception as a warning? I think this is a valid place to console.log something which we might be able to use for diagnostics later.

src/components/annotate/sidePanel/AnnotationForm.tsx
```diff
@@ -53,30 +54,45 @@ const AnnotationForm = (props: Props) => {
...
+            if (editAnnotation['oa:target']['oa:selector']?.['ods:field']) {
+                targetField = editAnnotation['oa:target']?.['oa:selector']?.['ods:field'] as string;
+            } else if (editAnnotation['oa:target']['oa:selector']?.['oa:class']) {
+                targetClass = editAnnotation['oa:target']?.['oa:selector']?.['oa:class'] as string;
+            }
+
             motivation = editAnnotation['oa:motivation'] as string;
         } else if (annotateTarget.targetProperty.type === 'field') {
             targetField = annotateTarget.targetProperty.name;
+            motivation = 'ods:adding';
```
samleeflang, line 73: It doesn't necessarily be an `ods:adding` annotation right? There could also be a comment annotation on a field or class.

src/components/annotate/sidePanel/AnnotationForm.tsx
```diff
@@ -102,14 +120,16 @@ const AnnotationForm = (props: Props) => {
 
             /* Set Class form fields */
             setClassFormFields(classForm.formFields);
+        } catch {
+            annotationValue = !isEmpty(editAnnotation) ? editAnnotation['oa:body']['oa:value'] : get(annotateTarget.target, targetField) as unknown[];
```
samleeflang, line 124: Same as the above.

src/components/annotate/sidePanel/SidePanel.tsx
```diff
@@ -110,14 +110,45 @@ const SidePanel = (props: Props) => {
         if (annotationFormToggle || !isEmpty(editAnnotation)) {
             dispatch(setAnnotationFormToggle(false));
             dispatch(setEditAnnotation({} as Annotation));
-        } else if (annotateTarget.targetProperty.name) {
-            ShowWithAllAnnotations();
+
+            if (annotateTarget.targetProperty.name && annotateTarget.targetProperty.type !== 'superClass') {
```
samleeflang, line 114: Is it really needed to call this a `superClass`? Or can the targetType just be empty with empty indicating that it is about the whole object?

samleeflang (commented):
Some small comments. Next time would be nice to have this in a separate PR, this one is a bit big.

src/components/annotate/sidePanel/AnnotationForm.tsx
```diff
@@ -77,7 +77,17 @@ const AnnotationForm = (props: Props) => {
 
         /* If annotating a class instance, grab all fields from schema */
         if (targetClass) {
-            const schemaProperties = ExtractFromSchema(targetClass.split('.').pop() as string);
+            /* Get last class of path by removing indexes and dot notations */
+            let classSchemaName: string = targetClass;
+
+            if (classSchemaName.endsWith(']')) {
```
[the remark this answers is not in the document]
samleeflang: Thanks for the explanation

src/components/annotate/sidePanel/Annotation.tsx
TomDijkema: Hmm not really sure its the right place. Basically it fails when the body does not contain a valid object to read, which is the case for annotations with only a string value for fields or with certain motivations like commenting etc. A class annotation, upon adding or editing its values, does have an internal object that can be read, so in that case the action will succeed. But there are also class annotations with just a value, like when commenting on it. The if else I had before only checked for if it was a class or not, then it would assume an object could be read if a class is there, but that is not the case. So that's why I chose to do the try catch, to always look if an object can be read from it, and otherwise use the annotation value present as a string. In this case, the catch block is not really solving an error, but rather using the string value instead of reading an object.

src/components/annotate/sidePanel/AnnotationForm.tsx
TomDijkema: Correct, this predefined motivation makes sure at least one is selected, otherwise the annotation form will not be shown (checks if a motivation is present), the user can change this any other motivation by using the dropdown (except for cases where adding or editing is enforced).

src/components/annotate/sidePanel/AnnotationForm.tsx
TomDijkema: Also the same explanation

src/components/annotate/sidePanel/SidePanel.tsx
TomDijkema: Well, I tried that at first, but I think I could not get the logic to work with each other, as there are loads of conditionals in the side panel, checking what to display. Also, this type is local variable, so its not ingested into the annotation object itself or something like that, I get its not the prettiest solution right now, but I rather fix another time, as I think the annotation form needs to be split up into more components to make it better.

src/app/Utilities.ts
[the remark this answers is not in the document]
samleeflang: :smile:

src/components/annotate/sidePanel/SidePanel.tsx
samleeflang: Ok maybe create a task for it too look again some time in the future

src/components/annotate/sidePanel/AnnotationForm.tsx
TomDijkema: Removed try catch, and replaced with if else statement

src/components/annotate/sidePanel/Annotation.tsx
TomDijkema: Removed try catch, and replaced with if else statement

src/components/annotate/sidePanel/SidePanel.tsx
TomDijkema: https://naturalis.atlassian.net/browse/DD-257?atlOrigin=eyJpIjoiZjRlNTA1ZTFmOTM2NGUwNDk4YmIyOGY3MGVmYTVmMTUiLCJwIjoiaiJ9

samleeflang (approved):
:1st_place_medal:

PR #85, opened 2024-01-26
Mas scheduling improvements

TomDijkema:
PR add features requested for the enhancement of the MAS and MAS Job Records.

Adds:
- Allow batch annotation checkbox, to indicate if batch annotations are allowed or not
- Filter for filtering on the Machine Job Record state
- Polling functionality for the Machine Job Record table, polls the API every 3 seconds to check for new or updated data
- Notification message that pops up if a scheduled MAS, that is remembered in the local state, has been completed

The polling and notification message functionality ideally has to be replaced with Web Sockets in the future.

Fixes:
- Bug with the view JSON button routing to the wrong address

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -322,8 +322,13 @@ p {
   cursor: pointer;
 }
 
+/* Form Components */
+.checkbox {
+  width: 1rem;
+  height: 1rem;
+}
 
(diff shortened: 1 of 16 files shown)


samleeflang (approved):
:sparkle:

PR #86, opened 2024-03-06
Mas fixes

TomDijkema:
Fixes for scheduling MAS.
Completion message should only appear with a successful POST.
Allow batching is now applied to every single MAS, instead of them all at the same time.

### DIFF
--- a/src/api/specimen/ScheduleSpecimenMAS.ts
+++ b/src/api/specimen/ScheduleSpecimenMAS.ts
@@ -30,7 +30,7 @@ const ScheduleSpecimenMAS = async (handle: string, MASRequest: Dict, token?: str
         } catch (error) {
             console.warn(error);
 
-            throw(error);
+            throw (error);
         }
     }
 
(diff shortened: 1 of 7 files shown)


src/api/specimen/ScheduleSpecimenMAS.ts
```diff
@@ -32,6 +29,8 @@ const ScheduleSpecimenMAS = async (handle: string, MASRequest: Dict, batching: b
             specimenMAS = data.data[0].attributes;
         } catch (error) {
             console.warn(error);
+
```
southeo, line 32: remove space

southeo (approved):
:cookie:

src/api/specimen/ScheduleSpecimenMAS.ts
TomDijkema: You mean the enter? That is intentional

southeo: alright well **I** would remove it :laughing: java conventions are different i guess

samleeflang (approved):

PR #90, opened 2024-03-29
Replace doi string and add warning message

TomDijkema:
Replace doi string and add warning message

### DIFF
--- a/src/components/demo/DOITooltip.css
+++ b/src/components/demo/DOITooltip.css
@@ -5,7 +5,7 @@
 }
 
 .tooltip {
-    z-index: 1000;
+    z-index: -1;
     background-color: white;
     opacity: 0;
     border: 1px solid lightgray;
@@ -22,6 +22,7 @@
 }
 
 .tooltip.active {
+    z-index: 1000;
     opacity: 1;
 }
 
@@ -43,6 +44,9 @@
 .widthRight {
     width: 80%;
     display: inline-block;
+    overflow: hidden;
+    white-space: nowrap;
+    text-overflow: ellipsis;
 }
 
 .digitalExtendedSpecimenTitle {
--- a/src/components/demo/DOITooltip.css
+++ b/src/components/demo/DOITooltip.css
@@ -84,3 +84,8 @@
 .DOILogo {
     width: 30px;
 }
+
+.warningMessage {
+    color: red;
+    padding-left: 10px;
+}
--- a/src/components/demo/DOITooltipDemo.tsx
+++ b/src/components/demo/DOITooltipDemo.tsx
@@ -38,7 +38,7 @@ const DOITooltipDemo = (props: Props) => {
     }
 
     /* Function to map Handle or DOI response to valid record */
-    const FormatResponse = (responseRecord: {[name: string]: any}) => {
+    const FormatResponse = (responseRecord: { [name: string]: any }) => {
         const record = {
             data: {
                 attributes: {
@@ -60,11 +60,7 @@ const DOITooltipDemo = (props: Props) => {
         try {
             /* If DOI does not contain local prefix, try to resolve with the general handler */
             if (doi.includes('TEST') || doi.includes('SANDBOX')) {
-                let environment: string = 'dev';
-
-                if (doi.includes('SANDBOX')) {
-                    environment = 'sandbox';
-                }
+                let environment: string = doi.includes('SANDBOX') ? 'sandbox' : 'dev';
 
                 const response = await fetch(`https://${environment}.dissco.tech/handle-manager/api/v1/pids/${doi.replace(process.env.REACT_APP_DOI_URL as string, '')}`);
                 const record = await response.json();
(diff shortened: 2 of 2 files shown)


src/components/demo/DOITooltip.css
```diff
@@ -5,7 +5,7 @@
 }
 
 .tooltip {
-    z-index: 1000;
+    z-index: -1;
```
southeo, line 8: Hardcoding?? :eyes:

src/components/demo/DOITooltipDemo.tsx
```diff
@@ -37,20 +37,68 @@ const DOITooltipDemo = (props: Props) => {
...
-            const response = await fetch(`https://dev.dissco.tech/handle-manager/api/v1/pids/${doi.replace(process.env.REACT_APP_DOI_URL as string, '')}`);
-            const record = await response.json();
+            /* If DOI does not contain local prefix, try to resolve with the general handler */
+            if (doi.includes('TEST') || doi.includes('SANDBOX')) {
+                let environment: string = 'dev';
 
-            if (record.data) {
-                setActive(true);
+                if (doi.includes('SANDBOX')) {
+                    environment = 'sandbox';
```
southeo, line 66: this is an awkward way to assign the environment var. Can you do something like: 
`let environment: doi.includes('TEST') ? string = 'dev' : string = 'sandbox'`

I think javascript uses the [conditional operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator) the same way as in java

src/components/demo/DOITooltipDemo.tsx
```diff
@@ -37,20 +37,68 @@ const DOITooltipDemo = (props: Props) => {
...
+                    referentName: responseRecord.values[16].data.value,
+                    specimenHost: responseRecord.values[18].data.value,
+                    specimenHostName: responseRecord.values[19].data.value,
+                    primarySpecimenObjectId: responseRecord.values[20].data.value,
+                    topicDiscipline: responseRecord.values[27].data.value,
+                    livingOrPreserved: responseRecord.values[29].data.value
+                }
+            }
+        };
+
```
southeo, line 54: remove extra space?

src/components/demo/DOITooltipDemo.tsx
```diff
@@ -37,20 +37,68 @@ const DOITooltipDemo = (props: Props) => {
...
+                    environment = 'sandbox';
+                }
+
+                const response = await fetch(`https://${environment}.dissco.tech/handle-manager/api/v1/pids/${doi.replace(process.env.REACT_APP_DOI_URL as string, '')}`);
+                const record = await response.json();
+
+                if (record.data) {
+                    setRecord(record);
+                }
+            } else if (doi.includes('20.5000.1025')) {
```
southeo, line 75: if the doi and handle apis have the same structure, maybe just use the handle api for 20.5000.1025 and 10.3535 PIDs?

src/components/demo/DOITooltipDemo.tsx
```diff
@@ -88,7 +136,7 @@ const DOITooltipDemo = (props: Props) => {
             <button type="button" style={{ color: 'blue', cursor: 'pointer' }} ref={targetRef} onClick={() => TriggerTooltip()}> {children} </button>
 
             <div id="disscoTooltip" className={`tooltip ${active && 'active'}`} ref={DOITooltipRef} style={offsetStyles}>
-                {record.data ?
+                {record?.data ?
```
southeo, line 139: ?

southeo: What does this do here

southeo (commented):
I got real comments!

src/components/demo/DOITooltip.css
TomDijkema: Not really, by default its now -1, the active class makes it 1000

src/components/demo/DOITooltipDemo.tsx
TomDijkema: True, in principal it does not affect any of the code game play, but in this case the one liner looks cleaner, so changed it

src/components/demo/DOITooltipDemo.tsx
TomDijkema: I like to have space (enter) in between a return statement and what is happening above if its on the same indent level

src/components/demo/DOITooltipDemo.tsx
TomDijkema: Was thinking about that, left it open because thought maybe there is going to be stuff outside of our reach (don't think so)

src/components/demo/DOITooltipDemo.tsx
TomDijkema: Normally you use to say 'ok, if this property is there, use this one', it can be undefined, you are not sure.
In this case I removed it again as it already is a check if the data property is there, so it would be duplicate.

src/components/demo/DOITooltipDemo.tsx
southeo: both the doi api and the handle api are managed externally. we only manage the dissco ones. 

Keep it like this if you want! just wanted to clarify

southeo (approved):
:pancakes:

PR #91, opened 2024-04-22
Dot notation and Keycloak token

TomDijkema:
Fixes the dot notation with an array when selecting from the class or properties select list, array did not append a dot, whilst that is necessary with JSON path.

Also tries to fetch the user information from the Keycloak parsed token.


samleeflang (approved):
:turtle:

PR #92, opened 2024-05-02
Improve doi tooltip

TomDijkema:
PR improves the DOI Tooltip based upon the new (second) design.

Also adds an API call to orchestration to GET MAS.
Has to be tested when the CORS regulations between disscover and the orchestration service have been fully allowed.


src/api/mas/GetMAS.ts
```diff
@@ -0,0 +1,35 @@
...
+import axios from 'axios';
+
+/* Import Types */
+import { Dict, JSONResult } from 'app/Types';
+
+
+const GetMAS = async (handle: string) => {
+    let machineAnnotationService = {} as Dict;
+
+    if (handle) {
```
southeo, line 11: what is this check for? if it's false, it looks like you still return `machineAnnotationService`, should you log an error here?

src/components/demo/DOITooltip.css
```diff
@@ -21,54 +25,92 @@
...
+}
+
 .tooltip.active {
     z-index: 1000;
     opacity: 1;
 }
 
+.tooltipBody {
+    width: 100%;
+    display: flex;
```
southeo, line 39: :muscle:

southeo (commented):
:label:

src/api/mas/GetMAS.ts
TomDijkema: It checks if a handle is provided, otherwise it won't make the call.
Yes I have to implement error handling here, that is exactly one of the improvements I am making in the refactoring process: error handling for fetch requests and returning exceptions :scream: 

I don't want to put too much effort into this now as I already changed the way of fetching in the first refactored code.
Will update this when I get to it in the refactoring process

southeo (approved):
:racehorse:

wouteraddink:
Hi Tom, the icons at the bottom are not yet linked: DES icon should link to: https://doi.org/10.1093/biosci/biac060 and the DOI icon should link to: ttps://www.doi.org/the-identifier/what-is-a-doi/. Also it would be nice if the DOI in the text is still clickable and directs you to the landingpage.

PR #93, opened 2024-05-30
Keycloak user details and cors reference to orchestration fix

TomDijkema:
Keycloak now returns user details to the front-end that may be used to predefine user information in the DiSSCo account upon the first login.
Also constructs urls from dev to dev-orchestration and sandbox to orchestration to resolve the CORS issue between disscover and the orchestration API.


samleeflang (approved):
:racehorse:

PR #102, opened 2024-09-10
Annotation wizard

TomDijkema:
Invokes the basics of the annotation wizard with the progress dots component to create a step wise process.


samleeflang (approved):
:mage_man:

southeo (approved):
# :mage_man:

PR #103, opened 2024-09-13
Annotation wizard select instance step

TomDijkema:
PR that adds the select instance step of the annotation wizard.

In this step the user will be able to set the instance to be annotated, based upon the previous selected annotation target (JSON path). This can be a new or existing instance. Existing instances will be show and can be selected. New instances are a bit more complicated in the sense that it needs to be checked if all parent classes are present before one can actually be added. For example: a Taxon Identification class, which lives as a child of the Identification class, cannot be added if there is not identification present. Thus an identification should be added first before an actual taxon identification can be added upon it. As a solution, the target will be set to identification which should generate a form including the child classes like taxon identification so the user can still simultaneously add them. Another case is that there are multiple identifications present and the user needs to define on which identification the taxon identification is going to be made. This is done using a dropdown that displays all available identifications. Choosing one will set the correct JSON path target for the child taxon identification class.

The parent check process is complicated and does a lot of checks, but should be done a valid manner to prevent broken chains in the digital object.

Selected instances, new or existing, will be visually different from others so the users knows what he has selected.


src/components/elements/annotationSidePanel/components/annotationWizard/steps/AnnotationSelectInstanceStep.tsx
```diff
@@ -0,0 +1,158 @@
...
+import NewInstance from './NewInstance';
+import ExistingInstance from './ExistingInstance';
+
+
+/* Props Type */
+type Props = {
+    superClass: DigitalSpecimen | DigitalMedia | Dict,
+    schemaTitle?: string,
+    formValues?: Dict,
+    SetFieldValue?: Function
```
southeo, line 27: Can this be made more specific?

src/components/elements/annotationSidePanel/components/annotationWizard/steps/NewInstance.tsx
```diff
@@ -0,0 +1,166 @@
...
+                    if (!parentNodes.length) {
+                        parentClasses.push({
+                            jsonPath: jp.stringify(pathArray).replaceAll('"', "'"),
+                            name: MakeJsonPathReadableString(jp.stringify(pathArray)),
+                            ...(index > 0 && { parentName: pathArray[index] }),
+                            present: false
+                        });
+                    }
+
+                    parentNodes.forEach((parentNode, parentIndex) => {
```
southeo, line 80: you have a nested loop here, do you think you could build parentClasses outside the loop?

southeo:
```
const newParentClasses = [];
parentNodes.forEach((parentNode, parentIndex) => {
    newParentClasses.push({
        /*...populate object...*/
    });
});
setParentClasses(newParentClasses)
```

src/components/elements/annotationSidePanel/components/annotationWizard/steps/NewInstance.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+    /* Hooks */
+    const trigger = useTrigger();
+
+    /* Base variables */
+    const [parentClasses, setParentClasses] = useState<ParentClass[]>([]);
+
+    /* Check for parent classes, they either contain the term 'has' or start with a capital */
+    trigger.SetTrigger(() => {
+        const parentClasses: ParentClass[] = [];
```
southeo, line 54: maybe move this to a helper function for readability

src/components/elements/annotationSidePanel/components/annotationWizard/steps/ParentClassification.tsx
```diff
@@ -0,0 +1,186 @@
...
+                                also adding the parent class.`
+                        }
+                    </p>
+                </div>
+
+                <Button type="button"
+                    variant="primary"
+                    disabled={selected}
+                    className="fs-5 mt-3 py-1 px-3"
+                    OnClick={() => {
```
southeo, line 84: `onClick()` ?

src/components/elements/annotationSidePanel/components/annotationWizard/steps/AnnotationSelectInstanceStep.tsx
TomDijkema: Its just the typing of the params, an explanation is provided in the function description.
formValues could be more specific, but its already an optional (but required) param due to how props are passed onto tab components. What would you like?

src/components/elements/annotationSidePanel/components/annotationWizard/steps/NewInstance.tsx
TomDijkema: There is a reason for this and that is because there can be multiple of the same parent, aka multiple events if you target the location class. Also, the JSON path selecter nodes function you see being used returns that level of of an array, containing indexes for all 'hits' (found indexes), which contains another array with all the instances themselves, so events for example. Hence the double loop

src/components/elements/annotationSidePanel/components/annotationWizard/steps/NewInstance.tsx
TomDijkema: Done

src/components/elements/annotationSidePanel/components/annotationWizard/steps/ParentClassification.tsx
TomDijkema: Normally yes, but this is a custom UI component of mine, and when I pass along functions to custom components I capitalize them to easily separate them from normal variables. So here it is passed as OnClick, but in the Button component it is assigned as onClick yes

src/components/elements/annotationSidePanel/components/annotationWizard/steps/AnnotationSelectInstanceStep.tsx
southeo: Can you be more specific with the `Function` type? Not sure if that's possible

southeo: In python you specify what the Dict is of, is that a thing in typescript? otherwise leave as is

southeo (approved):
:cactus:

PR #105, opened 2024-09-23
Annotation policy text

TomDijkema:
Add annotation policy text to annotation side panel.

Also adds a button to the annotation wizard that lets you exit it.

### DIFF
--- a/src/app/utilities/AnnotateUtilities.ts
+++ b/src/app/utilities/AnnotateUtilities.ts
@@ -2,16 +2,69 @@
 import jp from 'jsonpath';
 
 /* Import Utilities */
-import { ExtractFromSchema, ExtractClassesAndTermsFromSchema } from 'app/utilities/SchemaUtilities';
+import { ExtractFromSchema, ExtractClassesAndTermsFromSchema, MakeJsonPathReadableString } from 'app/utilities/SchemaUtilities';
 
 /* Import Types */
 import { DigitalSpecimen } from "app/types/DigitalSpecimen";
 import { DigitalMedia } from "app/types/DigitalMedia";
-import { AnnotationFormProperty, Dict } from "app/Types";
+import { AnnotationFormProperty, AnnotationTarget, ParentClass, Dict } from "app/Types";
 
 
 /* Utilities associated with annotating */
 
+
+/**
+ * Function to extract parent classes from a given super class and provide all their associated information
+ * @param params An object containing: the annotationTarget, superClass and parentClasses properties
+ * @returns The extracted parent classes
+ */
+const ExtractParentClasses = (params: {
+    annotationTarget: AnnotationTarget,
+    superClass: DigitalSpecimen | DigitalMedia | Dict
+}) => {
+    const { annotationTarget, superClass } = params;
+
+    /* Base variables */
+    const pathArray: string[] = ['$'];
+    const parents = jp.parse(annotationTarget.jsonPath).slice(1, -1);
+    const parentClasses: ParentClass[] = [];
+
+    /* For each parant, find and extract its class information */
+    parents.forEach((parent, index) => {
+        /* Filter path from schemas */
+        let parentName = parent.expression.value.replace('ods:', '').replace('dwc:', '').replace('dcterms:', '');
+
+        /* Add to pathArray */
+        pathArray.push(parent.expression.value);
+
+        if (parentName.includes('has') || parentName[0] === parentName[0].toUpperCase()) {
+            /* If a parent class is empty, it needs to be targeted, if it is not completely empty, check for indexes */
+            const parentNodes = jp.nodes(superClass, jp.stringify(pathArray).replaceAll('][', ']..[').replaceAll('"', "'"));
+
+            if (!parentNodes.length) {
+                parentClasses.push({
+                    jsonPath: jp.stringify(pathArray).replaceAll('"', "'"),
+                    name: MakeJsonPathReadableString(jp.stringify(pathArray)),
+                    ...(index > 0 && { parentName: pathArray[index] }),
+                    present: false
+                });
+            }
+
+            parentNodes.forEach((parentNode, parentIndex) => {
+                parentClasses.push({
+                    jsonPath: jp.stringify(parentNode.path).replaceAll('"', "'"),
+                    name: MakeJsonPathReadableString(jp.stringify(parentNode.path)),
+                    ...(index > 0 && { parentName: MakeJsonPathReadableString(pathArray[index]) }),
+                    present: (Array.isArray(parentNode.value) && !!parentNode.value.length) || (!Array.isArray(parentNode.value) && typeof (parentNode.value) === 'object'),
+                    ...(parentNode.value.length && { options: parentNode.value.length }),
+                    ...(parentClasses[parentIndex]?.options && { dependent: true })
+                });
+            });
+        }
+    });
+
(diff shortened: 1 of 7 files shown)


src/app/utilities/AnnotateUtilities.ts
```diff
@@ -2,17 +2,70 @@
...
+
+    /* Base variables */
+    const pathArray: string[] = ['$'];
+    const parents = jp.parse(annotationTarget.jsonPath).slice(1, -1);
+    const parentClasses: ParentClass[] = [];
+
+    /* For each parant, find and extract its class information */
+    parents.forEach((parent, index) => {
+        /* Filter path from schemas */
+        let parentName = parent.expression.value.replace('ods:', '').replace('dwc:', '').replace('dcterms:', '');
```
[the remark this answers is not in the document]
southeo: Handy in `MakeJsonPathReadableString` too

src/app/utilities/AnnotateUtilities.ts
```diff
@@ -2,17 +2,70 @@
...
+
+        /* Add to pathArray */
+        pathArray.push(parent.expression.value);
+
+        if (parentName.includes('has') || parentName[0] === parentName[0].toUpperCase()) {
+            /* If a parent class is empty, it needs to be targeted, if it is not completely empty, check for indexes */
+            const parentNodes = jp.nodes(superClass, jp.stringify(pathArray).replaceAll('][', ']..[').replaceAll('"', "'"));
```
southeo, line 42: `][', ']..[` is that correct? does `..` mean something special in javascript?

src/app/utilities/AnnotateUtilities.ts
```diff
@@ -2,17 +2,70 @@
...
+
+    /* For each parant, find and extract its class information */
+    parents.forEach((parent, index) => {
+        /* Filter path from schemas */
+        let parentName = parent.expression.value.replace('ods:', '').replace('dwc:', '').replace('dcterms:', '');
+
+        /* Add to pathArray */
+        pathArray.push(parent.expression.value);
+
+        if (parentName.includes('has') || parentName[0] === parentName[0].toUpperCase()) {
```
southeo, line 40: would parentName.substring(0,1) be better than parentName[0] ?

src/app/utilities/AnnotateUtilities.ts
```diff
@@ -2,17 +2,70 @@
...
+                    present: false
+                });
+            }
+
+            parentNodes.forEach((parentNode, parentIndex) => {
+                parentClasses.push({
+                    jsonPath: jp.stringify(parentNode.path).replaceAll('"', "'"),
+                    name: MakeJsonPathReadableString(jp.stringify(parentNode.path)),
+                    ...(index > 0 && { parentName: MakeJsonPathReadableString(pathArray[index]) }),
+                    present: (Array.isArray(parentNode.value) && !!parentNode.value.length) || (!Array.isArray(parentNode.value) && typeof (parentNode.value) === 'object'),
```
southeo, line 58: `typeof (parentNode.value) === 'object'` ... javascript doesn't have built-in values for this?

src/components/elements/annotationSidePanel/components/AnnotationPolicyText.tsx
```diff
@@ -0,0 +1,135 @@
...
+                    {/* Machine annotations */}
+                    <Row className="mt-2">
+                        <Col>
+                            <p className="fs-3 fw-bold">
+                                Machine Annotations
+                            </p>
+
+                            <ul>
+                                <li>
+                                    Machine agents and machine annotation services must undergo review and verification before being added to the Machine Annotation Service.
```
southeo, line 98: before being added to the Machine Annotation Service Registry?

src/components/elements/annotationSidePanel/components/AnnotationPolicyText.tsx
```diff
@@ -0,0 +1,135 @@
...
+                            <ul>
+                                <li>
+                                    Users can archive their own individual annotations.
+                                </li>
+                                <li>
+                                    We do not delete annotations. Archived annotations are still accessible via the API;
+                                    they are just not visible with the latest digital specimen version.
+                                </li>
+                                <li>
+                                    [add examples of archived annotation when we have one; we decided to use the term archive
```
southeo, line 119: probably shouldn't publish a todo on our site, even in test

southeo (commented):
:ocean:

src/app/utilities/AnnotateUtilities.ts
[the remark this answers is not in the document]
TomDijkema: Implemented it for `ExtractParents` function, `MakeJsonPathReadableString` seems not to work since there can be multiple of the same ods:, dwc: or dcterms:

src/app/utilities/AnnotateUtilities.ts
TomDijkema: That is something from the jsonPath library. When you provide two dots it will try to search and target any matching nodes within an object that align with the provided JSON path

src/app/utilities/AnnotateUtilities.ts
TomDijkema: It does not really matter. Getting it by index is just more straightforward than the subString method.
Often if it is about the first character every time and you know that, you can just use the first index, aka zero.
Substring can be useful for selecting a specific part of a string, or for example the last character because [-1] sadly does not work.
For the last character you can also use at(-1)

src/app/utilities/AnnotateUtilities.ts
TomDijkema: How do you mean? The object type is a built in type of JS, here I am checking if it is an object or not. Arrays also count as object types in JS

src/components/elements/annotationSidePanel/components/AnnotationPolicyText.tsx
TomDijkema: @sharifX

sharifX: yes, please remove this part.

src/components/elements/annotationSidePanel/components/AnnotationPolicyText.tsx
sharifX: yes. adding "registry" is a good idea. -> "Machine agents and machine annotation services must undergo review and verification before being added to the Machine Annotation Service Registry."

src/app/utilities/AnnotateUtilities.ts
southeo: :+1: nice

src/app/utilities/AnnotateUtilities.ts
southeo: sorry I meant it looks strange to hardcode the string "object" to check if parentNode's value is an object. seems like there should be a better way than to use string comparison

southeo: you can resolve this if there isn't a better way

southeo (approved):
few straggling comments, approving because they're not that serious but would be good if you replied before merging

src/app/utilities/AnnotateUtilities.ts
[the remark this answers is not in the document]
TomDijkema: Good call, added those in the latest PR!

PR #106, opened 2024-09-24
Add annotation summary step to annotation wizard

TomDijkema:
PR adds a fourth and final step to the annotation wizard that displays all of the values defined in the previous steps of the wizard like motivation, target and all of the annotation values. This step will allow the user to save the annotation, which will be implemented in the next PR.

Also includes a few fixes and optimizations for the annotation wizard.


southeo (approved):
small things, not blocking

PR #107, opened 2024-09-27
Process and Submit Annotations

TomDijkema:
PR adds the functionality to process and submit annotations to the API when a user clicks on the 'save' button at the end of the annotation wizard. The processing and creation of the annotation object are handled as utility functions. The form submit function posts the annotation towards the API and catches the result, going back to the annotation overview with the newly added annotation, or displays an error message when an error occurs during the connection with the API. A new exception has been added for failing POST requests.


southeo (approved):
looks good! I would suggest merging the parent branch to main first though then merging this with main after, that way the parent branch doesn't get too big.

PR #109, opened 2024-09-30
Search id card show message if media is not there

TomDijkema:
PR shows message in ID Card when media should be there, but is not there and add new boolean search filter to search filters menu.


southeo (approved):
:candy:

PR #110, opened 2024-09-30
Reintegrate the mobile interpreter

TomDijkema:
PR reintegrates the mobile interpreter. It is based on the type of browser the user is using, so no screen width. If a comprehensive mobile version is wanted in the future, this needs to be further defined as it needs its own best practices and guide lines for responsiveness and custom classes in app.css.


southeo (approved):
:no_mobile_phones:

PR #111, opened 2024-10-15
Nest sub classes in form

TomDijkema:
PR fixes the remaining issues within the annotation wizard form step, when defining actual values to be annotated.
Sub classes are now nested inside their parents, so the UI does not conflict with the way the data model is structured and thus does not allow the user for creating broken parent child links.

Fixes the issues with:
- Broken links being created due to child classes not being nested inside parent classes in the UI
- Local annotation value not being assigned correctly in the annotation values processing  (zero index issue)
- Child objects and terms of sub classes with arrays not appending the parent index, generating an invalid field name (path)


southeo (approved):
:bread:

PR #112, opened 2024-10-15
Super class only allow comments and assessments

TomDijkema:
PR applies the logic to only allow comments and assessment motivated annotations for super classes.
Also fixes two issues with annotating top level terms:
- Not being able to add one when one does not exist
- Having their default values (when editing) in the form field


southeo (approved):
:clinking_glasses:

PR #115, opened 2024-10-25
Replace repeated super class type definition with default super class…

TomDijkema:
… type in types.ts

Replace repeated super class type definition with default super class type in types.ts


southeo (approved):
:grapes:

PR #117, opened 2024-10-28
Refactor mas

TomDijkema:
PR refactors the MAS menu and integration.
MASs will now appear in the annotation side panel as a stand alone sub menu.
Running MASs and seeing them in the overview / filtering on state still works the same.

Polling MASs has been improved to also pick up the state filter changes to incorporate it in the request.


southeo (approved):
Looks good but I recommend renaming anything `MAS` to `Mas`, more consistent with the backend and I think naming conventions in general

PR #119, opened 2024-10-30
Use source system in local search

TomDijkema:
Use the source system identifier suffix in combination with the provided local identifier to search for physical specimen ID.

The query id is constructed as: localIdentifier:suffixSourceSystemId


southeo (approved):
:butter:

PR #121, opened 2024-11-06
Refactor visual annotation

TomDijkema:
PR refactors the visual annotation process and all of its components enhancing the code of it and smoothing the user experience.
Includes: showing visual annotations on the canvas, adding new, editing and deleting visual annotations.

It is currently not possible to edit the shape of the bounding box of an annotation as I could not yet find a method to implement this in a way it does not interfere with the other functionality.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,8 +8,8 @@
       "name": "disscover",
       "version": "0.3.0",
       "dependencies": {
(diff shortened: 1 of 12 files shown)


southeo (approved):
:100:  cant wait to see this in action

PR #123, opened 2024-11-08
Update specimen landing page

TomDijkema:
Update Digital Specimen landing page with updated properties, provided by the presentation.


southeo (approved):
Small code changes for big effect! :exploding_head:

PR #124, opened 2024-11-08
Add button that links to ORCID when it has not been linked yet

TomDijkema:
Add a button to the profile page for linking the ORCID identifier to an existing account within DiSSCo.


southeo (approved):
EFFICIENT.

samleeflang (approved):
Super!!!

PR #126, opened 2024-11-11
Refactor Home Tour

TomDijkema:
Refactor Home Tour.
Tweak some stuff about the old tour, but the essence stays the same.
Improved naming and variable passing, less global state used,


src/components/home/components/advancedSearch/AdvancedSearch.tsx
```diff
@@ -170,14 +180,16 @@ const AdvancedSearch = () => {
                                 <Button type="button"
                                     variant="blank"
                                     className="tc-accent py-0"
-                                    OnClick={() => setAdvancedSearchToggle(true)}
+                                    OnClick={() => SetAdvancedSearchToggle(true)}
```
southeo, line 183: should this not start with a lowercase letter?

southeo (approved):
:ice_cube:

src/components/home/components/advancedSearch/AdvancedSearch.tsx
TomDijkema: Yes if the function was defined in this file, which it was in the first place, but I moved it up and passed it as props.
One of the refactor rules I set is: passed functions will all receive a capital to separate them from normal variables, since functions are treated as equals to variables.

southeo (approved):
:cool:

PR #127, opened 2024-11-12
Refactor search tour

TomDijkema:
PR refactors the tours on the search page, including one about the page itself and one about comparing specimens.
Also reintroduces the selection menu for multiple tours on the same page and streamlines the process a bit.
Targets of the tour will now be specified with classes as there is no possibility to have multiple identifiers on HTML elements.

Introduces a new method in the useSearchFilters hook for resetting all search filters simultaneously.

### DIFF
--- a/src/app/Hooks.ts
+++ b/src/app/Hooks.ts
@@ -446,10 +446,24 @@ const useSearchFilters = () => {
         return searchParams.getAll(name).includes(value);
     };
 
+    /**
+     * Function to reset and remove all search filters
+     */
+    const ResetSearchFilters = () => {
+        [...searchParams.entries()].forEach(searchParam => {
+            console.log(searchParam);
+
+            searchParams.delete(searchParam[0]);
+        });
+
+        setSearchParams(searchParams);
+    };
+
     return {
         GetSearchFilters,
         SetSearchFilters,
-        CheckSearchFilter
+        CheckSearchFilter,
+        ResetSearchFilters
     };
 };
 
--- a/src/components/search/tourSteps/CompareTourSteps.tsx
+++ b/src/components/search/tourSteps/CompareTourSteps.tsx
@@ -1,0 +1,122 @@
+/* Import Dependencies */
+import { Steps } from 'intro.js-react';
+
+/* Import Config */
+import StepsConfig from 'app/config/StepsConfig';
+
+/* Import Hooks */
+import { useAppSelector, useAppDispatch, useSearchFilters } from 'app/Hooks';
+
+/* Import Store */
+import { getTourTopic, setTourTopic } from 'redux-store/GlobalSlice';
+import { getCompareDigitalSpecimen, setCompareDigitalSpecimen } from 'redux-store/SearchSlice';
+
+/* Import Types */
+import { DigitalSpecimen } from 'app/types/DigitalSpecimen';
+import { PaginationObject } from 'app/Types';
+
+/* Import Sources */
+import SearchTourStepsText from 'sources/tourText/search.json';
+
+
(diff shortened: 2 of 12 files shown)


src/app/Hooks.ts
```diff
@@ -446,10 +446,24 @@ const useSearchFilters = () => {
         return searchParams.getAll(name).includes(value);
     };
 
+    /**
+     * Function to reset and remove all search filters
+     */
+    const ResetSearchFilters = () => {
```
southeo, line 452: I think it would be more efficient if you returned a new empty list of search params instead of deleting every search param in a loop

src/components/search/tourSteps/CompareTourSteps.tsx
```diff
@@ -0,0 +1,122 @@
...
+                            q: 'bellis perennis'
+                        });
+
+                        dispatch(setCompareDigitalSpecimen([]));
+
+                        resolve();
+                    } else if (nextIndex === 3) {
+                        /* On step 4: simulate selecting specimen for comparison */
+                        if (!compareDigitalSpecimen?.length) {
+                            dispatch(setCompareDigitalSpecimen([pagination.records[0] as DigitalSpecimen]));
```
southeo, line 80: add optional chaining? 
`pagination.records?.[0] as DigitalSpecimen`

src/components/search/tourSteps/CompareTourSteps.tsx
```diff
@@ -0,0 +1,122 @@
...
+            intro: step,
+            element: `.tourCompare${index + 1}`
+        });
+    });
+
+    return (
+        <Steps enabled={tourTopic === 'compare'}
+            steps={steps}
+            initialStep={0}
+            onBeforeChange={(nextIndex) => {
```
southeo, line 62: maybe move this to a separate function?

src/components/search/tourSteps/CompareTourSteps.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+                            dispatch(setCompareDigitalSpecimen([...localCompareDigitalSpecimen]));
+
+                            resolve();
+                        } else {
+                            resolve();
+                        }
+                    } else if ([4, 5].includes(nextIndex)) {
+                        /* On step 5: simulate selecting of second specimen for comparison */
+                        const compareDigitalSpecimen: DigitalSpecimen[] = [];
```
southeo, line 95: I think this shadows a variable from outside scope

southeo (commented):
:boar:

src/app/Hooks.ts
TomDijkema: Would be nice, but unfortunately there is no function for that

TomDijkema: Also the reason why all the other functions in this hook do the deleting first, upon appending the search filters individually

src/components/search/tourSteps/CompareTourSteps.tsx
TomDijkema: Yes added, also for the other occurrences

src/components/search/tourSteps/CompareTourSteps.tsx
TomDijkema: Done, also did it for the other tours

src/components/search/tourSteps/CompareTourSteps.tsx
TomDijkema: Yes but that does not matter in this context, as it is in another scope than the compareDigitalSpecimen variable at the top

southeo (approved):

PR #128, opened 2024-11-12
Apply dynamic annotation cases

TomDijkema:
Apply dynamic annotation cases so that the digital media page renders its own and not those of the digital specimen page.


southeo (commented):
:boat:

southeo (approved):
:baby_chick:

PR #130, opened 2024-11-18
Implement open ds 0.4.0

TomDijkema:
Implements the newest version of Open Digital Specimen, version 0.4.0.
Changes all references to the previous version of the data model to the new one.

A bug is present in the edit annotation form which has been tried to resolve, without luck so far.
A separate issue has been made for it: https://naturalis.atlassian.net/browse/DD-1511?atlOrigin=eyJpIjoiMzE5ODczN2U0MTExNDA5ODk3OTg2MWI1Zjg0M2MzMDgiLCJwIjoiaiJ9.


southeo (commented):
i don't think you should use @id, 
dcterms:identifier -> for specimen/media
schema:identifier -> for everything else

Also, ignore the comment I made on the generated code

southeo (approved):

PR #131, opened 2024-11-19
Update digital media page content

TomDijkema:
Update the Digital Media page to be inline with the newest requirements (content wise).
Adds tabs to the page including one for the metadata and one for the entity relationships.

Also includes a few fixes for selecting a term instance when annotating (value did not appear for boolean values and term schemas without a type definition now show up by default in the select instance menu, similar to if the type is term).


southeo (approved):
:turtle:

PR #134, opened 2024-12-03
Add annotation mode

TomDijkema:
Adds the annotation mode to the Digital Specimen page to allow for direct targeting of properties from the view.
This includes some slight changes and fixes to the annotation wizard in general.

### DIFF
--- a/src/components/digitalSpecimen/DigitalSpecimen.tsx
+++ b/src/components/digitalSpecimen/DigitalSpecimen.tsx
@@ -1,6 +1,5 @@
 /* Import Dependencies */
 import classNames from 'classnames';
-import jp from 'jsonpath';
 import { useState } from 'react';
 import { Container, Row, Col } from 'react-bootstrap';
 import { useParams } from 'react-router-dom';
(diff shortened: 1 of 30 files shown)


djangobrunink (approved):
lgtm

src/components/digitalSpecimen/DigitalSpecimen.tsx
```diff
@@ -99,6 +100,19 @@ const DigitalSpecimen = () => {
         }
     });
 
+    /**
+     * Function to set the annotation target state
+     * @param annotationTargetType The type of the annotation target, either class or term
+     * @param jsonPath The JSON path that targets the class or term
+     */
+    const SetAnnotationTarget = (annotationTargetType: 'class' | 'term', jsonPath: string) => {
```
southeo, line 108: A lot of this seems duplicate from the media.tsx file, do you think you can reuse some code?

southeo (approved):
One comment but isn;t blocking, something to consider in a future pr
:coat:

src/components/digitalSpecimen/DigitalSpecimen.tsx
TomDijkema: Yeah my initial idea was to include as an utility function, only I am not allowed to use hooks out of React components, hence I chose to do it this way. I could include it at an ever higher level and pass it as props, but its irrelevant for a lot of child components. For now this seemed the best way, but can probably be optimized somehow in the future.

PR #136, opened 2024-12-06
Create data export page, add to navigation and handle data export req…

TomDijkema:
…uests

Create data export page, add to navigation and handle data export requests


src/api/dataExport/ScheduleDataExport.ts
```diff
@@ -0,0 +1,70 @@
...
+            data: {
+                type: 'export-job',
+                attributes: {
+                    searchParams: [
+                        {
+                            inputField: dataExportKey,
+                            inputValue: dataExportValue
+                        }
+                    ],
+                    targetType: 'https://doi.org/21.T11148/894b1e6cad57e921764e',
```
southeo, line 43: Can you pull this from another variable? I think you should have this somewhere

Also, `exportType` will eventually support a few options, might make more sense to store it as a variable.

src/components/dataExport/DataExport.tsx
```diff
@@ -0,0 +1,169 @@
...
+                                        dataExportValue: values.sourceSystemID
+                                    });
+
+                                    notification.Push({
+                                        key: `dataExport_${Math.random()}`,
+                                        message: 'Data Export scheduled successfully, your download link will be sent via email',
+                                        template: 'success'
+                                    });
+                                } catch {
+                                    notification.Push({
```
southeo, line 97: maybe put these push calls into a single function? 
```
const pushNotification = (message: string, template: 'success' | 'error') => {
    notification.Push({
        key: `dataExport_${Math.random()}`,
        message,
        template
    });
};
```

southeo (commented):
nice nice, just 2 small things

src/components/dataExport/DataExport.tsx
TomDijkema: The Push function already is the function you are describing here, so that would duplicate it.

src/api/dataExport/ScheduleDataExport.ts
TomDijkema: I added a dropdown for the exportType, but how do you mean pull the targetType from another variable? Its not directly linked to a digital specimen or something so I don't have anything to trace that down

TomDijkema: We will discuss this in a refinement because the fdo type can in the best case come from the schema

southeo (approved):
:cow:

PR #137, opened 2024-12-06
Implement function for retrieving environment variables

TomDijkema:
Implement function for retrieving environment variables


southeo (approved):
:raccoon:

PR #140, opened 2024-12-11
Prefill annotation classes with values

TomDijkema:
Introduce the functionality to prefill classes with values when annotating them.
This is only intended for annotations with the 'ods:adding' motivation.
The default values needs to be further refined, but the functionality should be in place.


src/app/utilities/AnnotateUtilities.ts
```diff
@@ -208,14 +209,33 @@ const GenerateAnnotationFormFieldProperties = async (jsonPath: string, superClas
...
+     * @param classValues Possible existing class values from the super class
+     * @param isArray Boolean indicating if the class represents an array
+     * @param localClassValues Possible existing local class values of an annotation
+     */
+    const DepictClassValues = (fieldName: string, classValues: Dict | Dict[] | undefined, isArray?: boolean, localClassValues?: Dict[]) => {
+        if (classValues) {
+            return classValues;
+        } else if (!isEmpty(localClassValues)) {
+            return localClassValues;
+        } else if (motivation !== 'oa:editing' && isArray) {
```
southeo, line 232: should this include 'oa:adding' ?

southeo (approved):
:frog:

src/app/utilities/AnnotateUtilities.ts
TomDijkema: Not this time. Prefill values is only allowed for the 'ods:adding' motivation.
Actually changed this check to be: motivation === 'ods:adding', to also exclude the other motivations.

PR #148, opened 2025-01-09
Minor fixes

TomDijkema:
Minor fixes that resolve the annotation form not appearing when editing class annotations.
And should now definitely fix the images on the acknowledgements page.


southeo (approved):
:camera_flash:

PR #151, opened 2025-03-20
Encode URI to get MAS job Record

southeo:
ORCID needs to be encoded in order to get mas job records for user


samleeflang (approved):
A frontend PR!!!! :window:

PR #155, opened 2025-03-26
update mjr type

southeo:


southeo:
Closing because we will wait for FE to come in

PR #157, opened 2025-05-27
Feature/mas job records

MelanieTheLion:
**In this PR**
- bugfix for the MasJobRecords table on the profile page breaking when one of the table cell values returns an undefined, in this case the 'completed' column. I've added a check on either a value or an undefined. In case of undefined, we set the value to '', to show that there is no value.

**Note**
Testing for these files currently not set up. I am working on a general testing setup that extends the already existing test setup

### DIFF
--- a/src/app/config/table/UserMasJobRecordsTableConfig.tsx
+++ b/src/app/config/table/UserMasJobRecordsTableConfig.tsx
@@ -19,8 +19,8 @@ const UserMasJobRecordsTableConfig = () => {
     /* User annotation record type */
     type UserMasJobRecord = {
         targetId: string,
-        scheduled?: string | undefined,
-        completed?: string | undefined,
+        scheduled: string | undefined,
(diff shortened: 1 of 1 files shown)


src/app/config/table/UserMasJobRecordsTableConfig.tsx
```diff
@@ -2,6 +2,14 @@
 import { createColumnHelper } from '@tanstack/react-table';
 import { format } from 'date-fns';
 
+/**
+ * Function to help deal with the date as a string or undefined
+ * @param date The time data either a string or undefined
+ * @returns Either an empty string or the formatted date
+ */
+function formatDate(date: string | undefined) {
```
JontySponselee-Naturalis, line 10: Optionally add a return type to make thee function more strict

JontySponselee-Naturalis (approved):

PR #158, opened 2025-05-30
align workflow file with other applications

southeo:
[Jira](https://naturalis.atlassian.net/browse/DD-1659?atlOrigin=eyJpIjoiZjM4MWE4NjVjZGU2NGJjMGEzMGU1NTJjMGI1NzYyYmMiLCJwIjoiaiJ9)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,24 +7,20 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        include:
-          - image: public.ecr.aws/dissco/${{ github.event.repository.name }}-production
-            keycloak_client: production-disscover-fe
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -7,24 +7,20 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    strategy:
```
samleeflang, line 10: Oef, the reason for this is because we need to create two FE images for the different environments.
As Tom wasn't able to inject an environmental variable into the FE on run we needed to switch the env's at build time.
That is why we got this structure where the VITE_KEYCLOAK_CLIENT is set on build.
It might be better to create two different build files and have them both be triggered

samleeflang (commented):
We need two images with different ENV's for the frontend

PR #159, opened 2025-06-03
feat: Added export description and adjusted syntax and styling a bit

MelanieTheLion:
What it now looks like:
Screenshot from 2025-06-04 11-26-25

- Adjusted styling a bit
- Added description and optional description
- Changed syntax to make it more semantic (only a bit)

**Note**
Sourcesystems is not properly mocked locally, only on dev and sandbox, so locally the second dropdown does not work at all.


JontySponselee-Naturalis (approved):

southeo (approved):
:tiger:

PR #164, opened 2025-06-25
Feature/add doi to specimen

MelanieTheLion:
Added:
- DOI to the Digital Specimen page in between the title and the MIDS level and version selector
- New tab called identifiers with the DOI and whatever is in 'ods:hasIdentifiers'

What it looks like:
image

### DIFF
--- a/src/app/config/table/IdentifiersTableConfig.tsx
+++ b/src/app/config/table/IdentifiersTableConfig.tsx
@@ -1,0 +1,34 @@
+/* Import Dependencies */
+import { createColumnHelper } from '@tanstack/react-table';
+
+/* Import Utilities */
+import { RetrieveEnvVariable } from 'app/Utilities';
+
+/**
+ * Config function that sets up the basic table column template for the search results table on the search page
+ * @returns Table columns
+ */
+const IdentifiersTableConfig = () => {
+    /* Search results type */
+    type Identifiers = {
+        DOI: string,
+    };
+
+    /* Base variables */
+    const columnHelper = createColumnHelper<Identifiers>();
+
+    /* Table columns */
+    const columns = [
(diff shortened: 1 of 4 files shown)


southeo (approved):
Looks good to me, I really like the Identifiers tab!

src/app/config/table/IdentifiersTableConfig.tsx
```diff
@@ -0,0 +1,39 @@
+/* Import Dependencies */
+import { createColumnHelper } from '@tanstack/react-table';
+
+/**
+ * Config function that sets up the basic table column template for the identifiers tab on the Digital Specimen page
+ * @returns Table columns
+ */
+const IdentifiersTableConfig = () => {
```
JontySponselee-Naturalis, line 8: Why is a function being used to get the config? At the moment there is no reason to recalculate `columns` whenever this function is invoked. It is a bit waste of compute, it could be simplified to a plain object.

src/components/digitalSpecimen/components/contentBlock/components/Identifiers.tsx
```diff
@@ -0,0 +1,67 @@
...
+    const { digitalSpecimen } = props;
+
+    /* Data table configuration */
+    const { columns } = IdentifiersTableConfig();
+
+    /* Set Table data with different identifiers */
+    const tableData: DataRow[] = [
+        {
+            key: 'DOI',
+            value: digitalSpecimen['@id'].replace(RetrieveEnvVariable('DOI_URL'), '')
```
JontySponselee-Naturalis, line 37: Is this a common thing to do, replacing `@id` or any `@` placeholder? Otherwise I would use utility function so you won't have typos for the possible options and will be easier to maintain if changes are needed

PR #165, opened 2025-06-25
feat: rewrite of annotation endpoint

MelanieTheLion:
Done:
- Removed the GetDigitalSpecimenAnnotations api service
- Replaced the result of GetDigitalSpecimenAnnotations service with result of GetDigitalSpecimenComplete
- Rewrote function in AnnotationsSidePanel to distinguish between digital media annotations and digital specimen annotations

Did not write any extra tests, have to do that later when the test suite will be handled.


src/components/elements/annotationSidePanel/AnnotationSidePanel.tsx
```diff
@@ -80,16 +80,20 @@ const AnnotationSidePanel = (props: Props) => {
...
-        },
-        triggers: [superClass, annotationWizardToggle],
-        Method: GetAnnotations,
-        Handler: (annotations: Annotation[]) => {
-            setAnnotations(annotations);
-        }
-    });
+    const handle = superClass?.['@id']?.replace(RetrieveEnvVariable('DOI_URL'), '');
+
+    if (handle && (superClass['@type'] === 'ods:DigitalSpecimen' || superClass['@type'] === 'ods:DigitalMedia')) {
```
southeo, line 85: Is it worth catching an error or logging a warning if `handle` is null? I think that would be good to know

southeo (approved):
:soap:

src/components/elements/annotationSidePanel/AnnotationSidePanel.tsx
JontySponselee-Naturalis: I second this. If you have plans to eventually add an error collecting tool then you can identify early in the userflow where the issue arises, because I assume the fetch beneath will fail and some error will be logged then? Will the user also be notified if there is an issue in both of these cases?

MelanieTheLion: Added two console.warns. This situation is not a forever situation. I want to implement error handling piece by piece, and in phases :).

southeo (approved):
Fair solution for now

PR #166, opened 2025-07-09
fix: Added a check to not confuse the tour dummy data with the annotation data

MelanieTheLion:
Crafty bug.

The tour dummy data was bleeding into the user annotation flow, due to state changes updating components. For now, I've added a safe guard to check whether the tour is active or not, and if not, do nothing. But this brings to light a bigger issue: two different ways of 'state management': props and the redux state, and components still in the DOM handling changes, while not active.

So I'll have to think about this, but since we are redesigning the annotation flow, this does not have any priority.


southeo (approved):
:mag_right:

samleeflang (approved):
Aaahhhh such a small thing.... 
Congrats on hunting it done :1st_place_medal:

PR #172, opened 2025-07-24
bug: Removed the flagsapi and flags images because it does not work with our data

MelanieTheLion:
Problem:
- Flagsapi expects the Iso alpha 2 country codes (ex.: NL), while we either get Iso alpha 2, alpha 3 (ex. NLD) or no country codes in our data, meaning that the flagsapi doesn't always work.

Solution:
- Instead of writing a condition to make it work with all three known cases, I chose to remove the flagsapi all together, because we are currently working with users. If we get feedback from them that they'd like to have a visual component relating to the country, we can revisit the idea and see if we can figure out a new solution.


JontySponselee-Naturalis (approved):

southeo (approved):
:no_entry_sign: :flags: :no_entry_sign:

PR #173, opened 2025-07-30
Feature/identification icon

MelanieTheLion:
In this PR:
- Added an icon to show the user that it can modify the Accepted Identification taxonomic data with an annotation
- The icon opens the correct annotation panel view
- The user cannot select a single term in the UI anymore
- The integration test and its mock data has been slightly added upon

Note:
- The current integration test setup will be rewritten to implement Playwright, and the piece of functionality is not done, so the current test is not complete.

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,7 +54,6 @@
         "@testing-library/jest-dom": "^6.6.3",
         "@testing-library/react": "^16.1.0",
         "@testing-library/user-event": "^14.5.2",
-        "@types/css-font-loading-module": "^0.0.14",
         "@types/cytoscape": "^3.21.8",
         "@types/jsonpath": "^0.2.4",
         "@types/leaflet": "^1.9.15",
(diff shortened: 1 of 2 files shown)


src/components/digitalSpecimen/components/contentBlock/components/DigitalSpecimenOverview.tsx
```diff
@@ -249,15 +250,33 @@ const DigitalSpecimenOverview = (props: Props) => {
...
+                                <Col lg="auto">
+                                    <Button type="button"
+                                        variant="blank"
+                                        className="px-0 py-0"
+                                        dataTestId="identification-annotation-button"
+                                        OnClick={() => {
+                                            // Toggle annotation mode
+                                            ToggleAnnotationMode();
+                                            // Set annotation target to taxonomic identification
+                                            SetAnnotationTarget('class', `$['ods:hasIdentifications'][${acceptedIdentificationIndex}]['ods:hasTaxonIdentifications'][0]`);
```
southeo, line 262: This might be something you need to set elsewhere but if this path doesn't exist in the specimen (e.g. there are no taxonIdentifications), the motiviation should be "ods:adding". Otherwise it should be "oa:editing".

southeo: You can create a story to check this, it doesn't need to be in this PR

MelanieTheLion: So like we checked: If there is no acceptedIdentification, the entire accepted Identification panel does not show up here. If a user wants to add a new identification, it is still going through the regular route :)

PR #180, opened 2025-08-25
bug: forget the digital specimen if I close search

MelanieTheLion:
Problem:
Whenever we use the search of disscover, and we select a digital specimen, that digital specimen is saved in the store of the app. This is expected behavior. However, when we return to the homepage or do something else in the application besides searching, selecting and viewing a digital specimen, that specific digital specimen that I selected earlier is still saved.

Solution:
I want to clean up the selected digital specimen when I go out of the search flow. I don't need to remember which specimen I selected earlier, if I am already in a different flow.

### DIFF
--- a/src/components/home/Home.tsx
+++ b/src/components/home/Home.tsx
@@ -21,6 +21,7 @@ import { Header, Footer } from 'components/elements/Elements';
 import { AdvancedSearch, DatasetDisclaimer, Introduction, SearchBar, TopicFilters } from './components/HomeComponents';
 import { Button } from 'components/elements/customUI/CustomUI';
 
+
 /**
  * Base component that renders the Home page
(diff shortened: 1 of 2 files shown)


src/components/home/Home.tsx
```diff
@@ -21,7 +21,6 @@ import { Header, Footer } from 'components/elements/Elements';
 import { AdvancedSearch, DatasetDisclaimer, Introduction, SearchBar, TopicFilters } from './components/HomeComponents';
 import { Button } from 'components/elements/customUI/CustomUI';
 
-
```
MelanieTheLion, line 24: By accident removed this, but I'll keep the change, because no one needs two white lines

southeo (approved):
:butterfly:

PR #181, opened 2025-08-26
bug: made the hasBody property in annotations optional, because deleting does not have a body

MelanieTheLion:
Problem:
Whenever a digital specimen has an annotation of motivation 'ods:deleting', the hasBody property is missing from the response. This breaks the frontend, since the interface of Annotation is expecting the hasBody property. With 'breaking' I mean that the frontend does not show you anything anymore.

Solution:
I kept the current functionality working by making the hasBody property for Annotations optional. If there is no hasBody, you can now still view the annotation, but it does not show you all the information (which makes sense, because a deleting annotation does not have more info to show you).

How it now looks:
image

### DIFF
--- a/src/app/Types.ts
+++ b/src/app/Types.ts
@@ -101,7 +101,7 @@ export type AnnotationTarget = {
     annotation?: {
         id: string,
         motivation: string,
-        values: string[]
+        values: string[] | undefined
     }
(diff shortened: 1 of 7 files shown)


southeo (approved):
:grapes:

PR #182, opened 2025-09-03
feat: Added catalog number if existent to the Digital Specimen Overview page

MelanieTheLion:
Users want to see not only the DOI of a digital specimen, but also the local identifier. In this case, we are showing the catalog number, if there is any. 

The way it looks:
image


southeo (approved):
:dizzy:

PR #189, opened 2025-09-11
feat: Added icon, annotation functionality and specific annotation config for georeference

MelanieTheLion:
In this PR:
- An icon that opens the georeference annotation flow when you click on it
- The georeference annotation flow only shows 6 georeference annotation fields
- A simple test to test this small piece of functionality
- Previous functionality made reusable :)

How it looks:
Screenshot from 2025-09-15 10-15-50

### DIFF
--- a/src/app/utilities/AnnotateUtilities.ts
+++ b/src/app/utilities/AnnotateUtilities.ts
@@ -456,7 +456,9 @@ const ReformatToAnnotoriousAnnotation = (annotation: Annotation, mediaUrl: strin
     return annotoriousAnnotation;
 };
 
-const AnnotationFormFields = (topic: string) : string[] => {
+type AnnotationTopicValues = `${AnnotationTopic}`
+
+const AnnotationFormFields = (topic: AnnotationTopicValues) : string[] => {
     switch (topic) {
         case AnnotationTopic.TaxonIdentification:
             return ['dwc:kingdom', 'dwc:phylum', 'dwc:class', 'dwc:order', 'dwc:family', 'dwc:genus', 'dwc:scientificName'];
@@ -489,7 +491,7 @@ const FilterAndReorderAnnotationProperties = (
     /* Set the form properties to expectedProperties if the user is trying to annotate either the Taxon Identification or Georeference */
     if (currentAnnotationClass.properties) {
         const props = currentAnnotationClass.properties.filter(prop =>
-            AnnotationFormFields(currentAnnotationClass.key)?.includes(prop.key)
+            AnnotationFormFields(currentAnnotationClass.key as AnnotationTopicValues)?.includes(prop.key)
         );
 
         if (currentAnnotationClass.key === AnnotationTopic.TaxonIdentification) {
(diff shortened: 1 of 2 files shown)


src/app/utilities/AnnotateUtilities.ts
```diff
@@ -452,12 +452,56 @@ const ReformatToAnnotoriousAnnotation = (annotation: Annotation, mediaUrl: strin
 };
 
 const AnnotationFormFields = (topic: string) : string[] => {
-    if (topic === 'taxonomy') {
-        return ['dwc:kingdom', 'dwc:phylum', 'dwc:class', 'dwc:order', 'dwc:family', 'dwc:genus', 'dwc:scientificName'];
-    };
-    return [];
+    switch (topic) {
```
JontySponselee-Naturalis, line 455: topic as a string can lead to bugs by typos or cases that are missed, you can use an union or an enum to reduce the risk of this.

I think I said the same thing for return type of this, more and more code is depending on the specific annotation name, the risk of resulting into bugs becomes higher if you write more functionality around it

src/tests/components/digitalSpecimen/DigitalSpecimen.test.tsx
```diff
@@ -51,4 +51,19 @@ describe("General Digital Specimen Page Tests", () => {
         // Check if annotation panel is visible
         expect(annotationTarget).toBeVisible();
     });
+
+    /* Test if annotation panel opens when clicking on the georeferenceAnnotationButton */
+    it('opens the annotation panel when clicking the button', async () => {
+        // Find button
+        const annotationButton = await screen.findByTestId('georeference-annotation-button');
```
JontySponselee-Naturalis, line 58: When writing tests with testing-library/react try to not use a testID, but to use what users see like using the role of an element and text. This will also promote better A11Y, because you implicitly test for it

MelanieTheLion: Adjusted to using the role of button and added accessible test to the icon.
I also wrapped the button in the Tooltip, instead of the tooltip in the button, because this also turned out to be more accessible :)

src/app/utilities/AnnotateUtilities.ts
MelanieTheLion: Please take a look now. I added a string enum. The value of that enum ultimately comes from the data.

JontySponselee-Naturalis: Nice, I would also change the type for the parameter of this function, then passing strings to this function will also become more robust. If you don't want to use the enum to pass the string, an alternative is to create a type for the values

```ts
type AnnotationTopicValues= `${AnnotationTopic}`;
```

MelanieTheLion: Can you explain this a bit more?

JontySponselee-Naturalis:
```ts
const AnnotationFormFields = (topic: string) : string[] => {
  ....
}
```

The parameter `topic` is still a `string` here so you can pass any string to the function, e.g.

```ts
// Allowed by compiler
AnnotationFormFields("foo");

// Allowed by compiler
AnnotationFormFields("TaxonIdentification")
```

But to give some better developer experience you can use the enum as type, e.g.

 ```ts
const AnnotationFormFields = (topic: AnnotationTopic) : string[] => {
  ....
}

// Disallowed by compiler
AnnotationFormFields("foo");

// Disallowed by compiler
AnnotationFormFields("TaxonIdentification")

// Disallowed by compiler
AnnotationFormFields(AnnotationTopic.Foo)

// Allowed by compiler
AnnotationFormFields(AnnotationTopic.TaxonIdentification)
```

However in some cases you can't work with the enum unless you parse it to the correct type, but instead you can use an union of the values of the enum.

 ```ts
// Will be `TaxonIdentification | Georeference`
type AnnotationTopicValues= `${AnnotationTopic}`; 

const AnnotationFormFields = (topic: AnnotationTopicValues) : string[] => {
  ....
}

// Disallowed by compiler
AnnotationFormFields("foo");

// Allowed by compiler
AnnotationFormFields("TaxonIdentification")

// Allowed by compiler
AnnotationFormFields(AnnotationTopic.TaxonIdentification)
```

JontySponselee-Naturalis (approved):

PR #199, opened 2025-10-28
feat: Mapped data, description and titles

MelanieTheLion:
In this PR:
- Adjusted the titles of the classes to be a bit more human-friendly and readable.
- Removed some keys from the data to show in the annotation cards
- Only show adjusted fields in the annotation summary

Not in, but wip:
To not allow an 'empty' annotation to happen (meaning any annotation that does not have any changed fields).
image
image

### DIFF
--- a/src/app/utilities/tests/SchemaUtilities.test.ts
+++ b/src/app/utilities/tests/SchemaUtilities.test.ts
@@ -8,42 +8,23 @@ import { ExtractLastSegmentFromPath } from '../SchemaUtilities';
  * Description of tests to validate the functionality in Schema Utilities
  */
 describe("the ExtractLastSegmentFromPath functionality", () => {
-    let jsonPath: string;
-    let result: string | undefined;
-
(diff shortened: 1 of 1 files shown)


src/app/utilities/SchemaUtilities.ts
```diff
@@ -271,9 +271,35 @@ const StripSchemaString = (jsonPath: string) => {
...
+    /* If jsonPath is root, return generic name */
+    if (jsonPath === '$') return 'Digital Specimen';
+
+    /* Split the path by the '[' character to get parts */
+    const parts = jsonPath.split('[');
+
+    /* Find the last part that is not just a number (to ignore array indices) */
+    const lastMeaningfulPart = [...parts].reverse().find(part => !/^\d+\]$/.test(part));
+
+    if (!lastMeaningfulPart) return '';
```
southeo: maybe return whole jsonPath instead?

JontySponselee-Naturalis (approved):

PR #200, opened 2025-11-03
Feature/annotation UI changes

MelanieTheLion:
In this PR:
- Added 'annotate' to the annotation icon for georeference and taxon identification
- Showing only the adjusted values in the summary
- Updated dropdown behavior to something more intuitive

image

### DIFF
--- a/src/components/digitalSpecimen/components/contentBlock/components/DigitalSpecimenOverview.tsx
+++ b/src/components/digitalSpecimen/components/contentBlock/components/DigitalSpecimenOverview.tsx
@@ -213,7 +213,6 @@ const DigitalSpecimenOverview = (props: Props) => {
                                             Annotate
                                             <FontAwesomeIcon icon={faPenToSquare}
                                                 className="ps-2"
-                                                title="Add an annotation to modify the geological reference"
                                             />
                                         </span>
                                     </Tooltip>


src/components/digitalSpecimen/components/contentBlock/components/DigitalSpecimenOverview.tsx
```diff
@@ -210,10 +210,13 @@ const DigitalSpecimenOverview = (props: Props) => {
...
                                     >
-                                        <FontAwesomeIcon icon={faPenToSquare}
-                                            className="tc-grey"
-                                            title="Add an annotation to modify the geological reference"
-                                        />
+                                        <span className="tc-primary fs-5 fw-lightBold">
+                                            Annotate
+                                            <FontAwesomeIcon icon={faPenToSquare}
+                                                className="ps-2"
+                                                title="Add an annotation to modify the geological reference"
```
JontySponselee-Naturalis: I don't think title is needed here since the icon is helping the button text, it is not a standalone element. However, I do think the text makes it more clear what the functionality actually does: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#icon_buttons

I also see that the button (not changed in this MR) has a `aria-label` while this is not needed when it already has text:
https://developer.mozilla.org/en-US/docs/Glossary/Accessible_name

I am not sure about the best implementation, but I think the text that you used for the title makes more sense as a tooltip for the button (instead of the icon).

src/components/digitalSpecimen/components/contentBlock/components/DigitalSpecimenOverview.tsx
```diff
@@ -197,28 +197,27 @@ const DigitalSpecimenOverview = (props: Props) => {
...
+                                    variant="blank"
+                                    className="px-0 py-0"
+                                    OnClick={() => {
+                                        // Toggle annotation mode
+                                        ToggleAnnotationMode();
+                                        // Set annotation target to geological reference
+                                        SetAnnotationTarget('class', `$['ods:hasEvents'][0]['ods:hasLocation']['ods:hasGeoreference']`);
+                                    }}
+                                >
+                                    <Tooltip text="Add an annotation to modify the geological reference" placement="bottom">
```
JontySponselee-Naturalis: This is better, but I think that the button needs `aria-describedby` if you want to be complete, not entirely sure though

MelanieTheLion: Not sure about that, aria-describedby seems to reference id's and I am not using id's. In this case, the 'annotate' of the button text would be the accessible element.

src/components/digitalSpecimen/components/contentBlock/components/DigitalSpecimenOverview.tsx
```diff
@@ -197,28 +197,27 @@ const DigitalSpecimenOverview = (props: Props) => {
...
                                         <span className="tc-primary fs-5 fw-lightBold">
                                             Annotate
                                             <FontAwesomeIcon icon={faPenToSquare}
                                                 className="ps-2"
                                                 title="Add an annotation to modify the geological reference"
```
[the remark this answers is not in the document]
JontySponselee-Naturalis: No need for apologies, I am just pointing out inconsistencies that you perhaps overlooked :P

PR #203, opened 2025-11-12
Various improvements

MelanieTheLion:
Removed an unnecessary class and its usage, since it resulted in weird formatting.
Adjusted the size of the content boxes to not have content disappear.


southeo (approved):
:swan:

PR #204, opened 2025-11-19
feat: Replaced some aggregation filters with hardcoded filters

MelanieTheLion:
The following search filters are now hardcoded and not dependent on the aggregation api :)
- hasMedia
- LivingOrPreserved
- missingData
- MidsLevel

This is a break from the working functionality as it was, and thus means quite some changes to work this in.
There is no screenshot, because this is not a UI change

### DIFF
--- a/src/app/Types.ts
+++ b/src/app/Types.ts
@@ -71,7 +71,7 @@ export type SearchFilter = {
     searchable?: boolean,
     searchAlias?: string,
     filters?: string[],
-    noAggregations?: boolean,
+    noAggregations?: boolean
 };
 
 /* Search filters (fetch requests) */


src/components/search/components/searchFiltersMenu/components/BlockFilter.tsx
```diff
@@ -26,16 +28,18 @@ type Props = {
...
 
     /* Base variables */
     const blockFilterItems: string[] = [];
 
     /* Construct from boot aggregations */
-    if (aggregations) {
-        Object.keys(aggregations).forEach(key => {
-            blockFilterItems.push(key);
-        })
+    if (aggregations && !noAggregations) {
```
southeo: is this redundant?

PR #211, opened 2025-12-04
feat: Removed the tour and test suite

MelanieTheLion:
**In this PR:**
I've removed the tour functionality, because we don't want it right now and it makes the application incredibly complicated.
I've also removed the current test suite and its tests and the test step in our build file. I'll be adding and rewriting this in due time, but for now the test setup is inadequate and instead of 'fixing' it, I'd like to build it up from scratch. **[TEMPORARY]**
I have also removed an api service that we are not using and that I don't have to use.

**Tests done to validate that the app still works:**
- Search a specimen
- Create an annotation
- Delete an annotation
- Run a mass
- Check virtual collections

Visually the only thing different is that the tour button is gone and the mobile screen is gone:
image


southeo (approved):
I'm happy. I also appreciate the note on what functional tests you did to make sure this doesn't have unintended consequences :)

PR #216, opened 2025-12-22
Refactoring header

MelanieTheLion:
**In this PR:**
- New header according to the design linked below
- Mobile version of the header
- max-width on the header, but not on the rest of the content (couldn't make it work without having to do a LOT of changes)
- Styling sheet for the header

**Accessibility:**
- checked with the Wave A11Y browser extension and Google Lighthouse. No issues currently in the new header (but ignoring the rest of the pages)
image
- I am able to navigate via keyboard through my menu and take the actions that I can take.

**Not included:**
- Testing suite for unit tests
- And unit tests for the Header specific
I want to separate this from this PR, because it will change some vite config as well.

What the header and the app look like now:
image

Mobile:
image


src/components/elements/header/Header.scss
```diff
@@ -0,0 +1,59 @@
...
+            display: none;
+        }
+
+        .nav-buttons {
+            height: 100%
+        }
+    }
+    a {
+        #text-logo-disscover {
+            font-size: 32px;
```
JontySponselee-Naturalis: Same comment as for padding to standardize the different font-sizes and font-weight

JontySponselee-Naturalis (approved):

PR #244, opened 2026-04-03
Refactor of GetDigitalSpecimenComplete service and setup of data mapper to UI

MelanieTheLion:
In this PR:
- Rewrote service getDigitalSpecimenComplete to use tanstack and execute dataMapper for digital specimen
- Added data mapper function to map data to UI in one general place, instead of multiple places in the app

This service and data mapper is not used yet, but is tested. Implementation to UI will follow this PR due to complexity.

### DIFF
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -22,3 +22,6 @@ CVE-2026-33671
 # Vulnerability in image dependency libpng that is included in the image and we cannot fix Apr-1-2026
 CVE-2026-33416
 CVE-2026-33636
+
+# lodash-es is pinned as a dependency on a vulnerable version in formik and cannot be upgraded by us Apr-3-2026
+CVE-2026-4800
--- a/src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
+++ b/src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
@@ -1,0 +1,40 @@
+import apiClient from '../apiClient';
+
+/**
+ * Service that retrieves the complete digital specimen details through the apiClient
+ * Takes handle and version as a parameter object
+ * @returns An array with complete digital specimen data on specimen, media and annotations
+ */
+export const getDigitalSpecimenComplete = async ({ handle, version }:
+    { handle: string, version?: number }) => {
+    let endPoint: string;
+
+    if (version) {
+        endPoint = `digital-specimen/v1/${handle}/${version}/full`;
+    } else {
+        endPoint = `digital-specimen/v1/${handle}/full`;
+    }
+    try {
+        /* Call service and wait for response */
+        const response = await apiClient.get(endPoint, {
+            params: {
+                handle,
+                version
+            }
+        });
+
+        /* Throw error if response is not as expected */
+        if(!response.data?.data) {
+            throw new Error('Incorrect response format');
+        }
+
+        /* Return response data */
+        return response.data;
+    } catch (error) {
+        /* If error, throw error with generic error message */
+        console.error('Error fetching the digital specimen:', error);
+
+        /* Rethrow error for useQuery */
+        throw error;
+    }
+}
--- a/src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
+++ b/src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
@@ -2,23 +2,23 @@ import apiClient from '../apiClient';
 
 /**
  * Service that retrieves the complete digital specimen details through the apiClient
- * Takes handle and version as a parameter object
+ * Takes doi and version as a parameter object
  * @returns An array with complete digital specimen data on specimen, media and annotations
  */
-export const getDigitalSpecimenComplete = async ({ handle, version }:
-    { handle: string, version?: number }) => {
+export const getDigitalSpecimenComplete = async ({ doi, version }:
+    { doi: string, version?: number }) => {
     let endPoint: string;
 
     if (version) {
-        endPoint = `digital-specimen/v1/${handle}/${version}/full`;
+        endPoint = `digital-specimen/v1/${doi}/${version}/full`;
     } else {
-        endPoint = `digital-specimen/v1/${handle}/full`;
+        endPoint = `digital-specimen/v1/${doi}/full`;
     }
     try {
         /* Call service and wait for response */
         const response = await apiClient.get(endPoint, {
             params: {
-                handle,
+                doi,
                 version
             }
         });
--- a/src/utils/DataMappers/digitalSpecimenDataMapper.ts
+++ b/src/utils/DataMappers/digitalSpecimenDataMapper.ts
@@ -23,7 +23,7 @@ const getAcceptedIdentification = (ds: any) => {
 /**
  * Function to retrieve the primary event once
  * @param ds The digital specimen object
- * @returns Either the primary event or null
+ * @returns Either the primary event if there is one or null
  */
 const getPrimaryEvent = (ds: any) => {
     return ds["ods:hasEvents"]?.[0] ?? null;
--- a/src/utils/DataMappers/digitalSpecimenDataMapper.ts
+++ b/src/utils/DataMappers/digitalSpecimenDataMapper.ts
@@ -1,21 +1,8 @@
 /* Import schemas */
 import DIGITAL_SPECIMEN_SCHEMA_MAP from "./schemas/DigitalSpecimenSchema";
 
-/* Define interfaces */
-interface UIProperty {
-    label: string;
-    value: any;
-    isHtml: boolean;
-    type: string;
-}
-
-interface DigitalSpecimenUIModel {
-    SPECIMEN_RECORD: Record<string, UIProperty>;
-    IDENTIFICATION: Record<string, UIProperty>;
-    LOCATION: Record<string, UIProperty>;
-    COLLECTING_EVENT: Record<string, UIProperty>;
-    CITATION_LICENSE: Record<string, UIProperty>;
-}
+/* Import types */
+import { DigitalSpecimenUIModel, SchemaMap, UIProperty } from "./types/dataMapperTypes";
 
 /**
  * Transforms raw Digital Specimen data into a UI-ready model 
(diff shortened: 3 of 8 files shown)


src/utils/DataMappers/digitalSpecimenDataMapper.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/* Import types */
+import { DigitalSpecimenUIModel, SchemaMap, UIProperty } from "./types/dataMapperTypes";
+
+/**
+ * Function to get the accepted identification or the first one it can find
+ * @param ds The digital specimen object
+ * @returns Either the accepted identification or, if there is none, the first identification it can find
+ */
+const getAcceptedIdentification = (ds: any) => {
```
JontySponselee-Naturalis, line 12: This can return `undefined` when `identifications` is falsy, is this expected? Because the `@returns` suggests it will always return an identification

src/utils/DataMappers/digitalSpecimenDataMapper.ts
```diff
@@ -0,0 +1,47 @@
...
+    const acceptedIdentification = getAcceptedIdentification(ds);
+
+    return Object.entries(DIGITAL_SPECIMEN_SCHEMA_MAP as SchemaMap).reduce((accumulator, [groupKey, fields]) => {
+        const mappedFields: Record<string, UIProperty> = {};
+
+        Object.entries(fields).forEach(([fieldKey, config]) => {
+            mappedFields[fieldKey] = {
+                label: config.label,
+                value: config.resolve(ds, acceptedIdentification),
+                isHtml: !!config.isHtml,
```
JontySponselee-Naturalis: I know `!!` is a quite a common trick, but it is not explicit, `!Boolean(config.isHtml)`, makes it more clear that it converts it to an boolean and invert it

src/utils/DataMappers/digitalSpecimenDataMapper.ts
```diff
@@ -0,0 +1,47 @@
...
+    const ds = rawData?.data?.attributes?.digitalSpecimen;
+    if (!ds) return null;
+
+    /* Find acceptedIdentification or second best option */
+    const acceptedIdentification = getAcceptedIdentification(ds);
+
+    return Object.entries(DIGITAL_SPECIMEN_SCHEMA_MAP as SchemaMap).reduce((accumulator, [groupKey, fields]) => {
+        const mappedFields: Record<string, UIProperty> = {};
+
+        Object.entries(fields).forEach(([fieldKey, config]) => {
```
JontySponselee-Naturalis: You could make your code a bit more compact and immutable, but this is optional and opinionated.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries#object_transformations

src/utils/DataMappers/digitalSpecimenDataMapper.ts
```diff
@@ -0,0 +1,47 @@
...
+                label: config.label,
+                value: config.resolve(ds, acceptedIdentification),
+                isHtml: !!config.isHtml,
+                type: config.type || 'base',
+            };
+        });
+
+        accumulator[groupKey as keyof DigitalSpecimenUIModel] = mappedFields;
+        return accumulator;
+    }, {} as DigitalSpecimenUIModel);
```
JontySponselee-Naturalis: `.reduce` accepts a generic what it should return, this is more typesafe

src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
```diff
@@ -0,0 +1,40 @@
+import apiClient from '../apiClient';
+
+/**
+ * Service that retrieves the complete digital specimen details through the apiClient
+ * Takes handle and version as a parameter object
+ * @returns An array with complete digital specimen data on specimen, media and annotations
+ */
+export const getDigitalSpecimenComplete = async ({ handle, version }:
+    { handle: string, version?: number }) => {
```
southeo: nitpick but this is a doi not handle

src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
```diff
@@ -0,0 +1,40 @@
...
+        /* Call service and wait for response */
+        const response = await apiClient.get(endPoint, {
+            params: {
+                handle,
+                version
+            }
+        });
+
+        /* Throw error if response is not as expected */
+        if(!response.data?.data) {
```
southeo, line 27: to confirm: this will throw an error if you get a json that doesn't match the digital specimen structure? 

So if we get:

```
{
  "errorMsg" : "smth broke :("
}

```
as a response, will we fail here?

src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        /* Return response data */
+        return response.data;
+    } catch (error) {
+        /* If error, throw error with generic error message */
+        console.error('Error fetching the digital specimen:', error);
+
+        /* Rethrow error for useQuery */
+        throw error;
```
southeo, line 38: why catch here just to log it? why not just add the logging on line 28?

src/utils/DataMappers/schemas/DigitalSpecimenSchema.ts
```diff
@@ -0,0 +1,143 @@
...
+        discipline: {
+            label: 'Discipline',
+            resolve: (ds) => ds["dwc:topicDiscipline"]
+        }
+    },
+  
+    IDENTIFICATION: {
+        scientificName: {
+            label: 'Scientific Name',
+            isHtml: true,
```
southeo: `acceptedIdentification?.["dwc:scientificName"]` might not be html, so should this be set according to whether or not you've retrieved the html label or not?

src/utils/DataMappers/schemas/DigitalSpecimenSchema.ts
```diff
@@ -0,0 +1,143 @@
...
+        species: {
+            label: 'Species',
+            resolve: (_, { acceptedIdentification }) => acceptedIdentification?.["dwc:species"]
+        },
+        specificEpithet: {
+            label: 'Specific Epithet',
+            resolve: (_, { acceptedIdentification }) => acceptedIdentification?.["dwc:specificEpithet"]
+        },
+        nomenClaturalCode: {
+            label: 'Nomen/Clatural Code',
```
southeo: Is this a typo?

src/utils/DataMappers/types/dataMapperTypes.ts
```diff
@@ -0,0 +1,38 @@
+/* MapperContext to set the context of the data, e.g. acceptedIdentification */
+interface MapperContext {
+    acceptedIdentification?: any;
+    primaryEvent?: any;
+}
+
+/* UI Property interface to map data to */
```
southeo: missing end of sentence?

southeo (commented):
🐝

src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
MelanieTheLion: In the future, this will be caught by some form of error notification system I still have to implement. The error in the catch can be any kind of error, 500s, 404s. The error on line 28 is only thrown if the status is OK (200, 201), but the response that we get does not match what we expect at all. 
It is an edge case, because if it is OK, your response would probably (assumption) be what we expect.

The catch takes care of my error codes. So separate error instances.

src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
MelanieTheLion: Yes, as long as your response is OK (200, 201). 
Once your status code is an error code (400, 401, 404, 500 etc), it will be caught by the catch block and throw a different error.

So status code 200 and response is your example: I cannot process it, because it is not what I need and expect.
I throw error, Catch block will catch that error in response and let me know something went horribly wrong.

src/utils/DataMappers/digitalSpecimenDataMapper.ts
MelanieTheLion: I adjusted the function to also return a null in case of no identifications whatsoever. Probably a bit of an edge case, but my mapper works also if no identifications are found.

src/utils/DataMappers/digitalSpecimenDataMapper.ts
MelanieTheLion: Changed it to Boolean(config.isHtml). Turned out to be the right one :)

src/utils/DataMappers/digitalSpecimenDataMapper.ts
MelanieTheLion: I like this approach! I addressed it as well as I could. Let me know what you think.

src/utils/DataMappers/digitalSpecimenDataMapper.ts
MelanieTheLion: What it should return added.

src/services/digitalSpecimenService/getDigitalSpecimenComplete.ts
MelanieTheLion: Changed! Shameless copy from what was already there.

src/utils/DataMappers/schemas/DigitalSpecimenSchema.ts
MelanieTheLion: I added a check now to see which one I will get back. It is probably a bit overkill, but I think your point is valid. Transforming strings to HTML without any check is considered a bad practice.

src/utils/DataMappers/schemas/DigitalSpecimenSchema.ts
MelanieTheLion: Haha yes. Sometimes I don't know what I am talking about

src/utils/DataMappers/types/dataMapperTypes.ts
MelanieTheLion: No, just a weird structure. Removed the 'to'

southeo (approved):
🎏

.github/workflows/.trivyignore
```diff
@@ -21,4 +21,7 @@ CVE-2026-33671
 
 # Vulnerability in image dependency libpng that is included in the image and we cannot fix Apr-1-2026
 CVE-2026-33416
-CVE-2026-33636
\ No newline at end of file
+CVE-2026-33636
+
+# lodash-es is pinned as a dependency on a vulnerable version in formik and cannot be upgraded by us Apr-3-2026
```
JontySponselee-Naturalis, line 26: are you sure this cannot be changed with `overrides` ? Or are all the earlier packages vulnerable and new versions not compatible with formik ? 
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides

MelanieTheLion: I think I could. I have a follow up user story to address all these trivyignores, because the list is getting long and I'd like to clean them up.

JontySponselee-Naturalis (approved):

PR #246, opened 2026-04-20
Refactor of Digital Specimen overview components

MelanieTheLion:
**In this PR:**
- Created first set up with the Digital Specimen cards for the overview page with their respective data
- Cards will not be shown if there is no data for that specific subject
- The hero with the specific info on the digital specimen

**Not in this PR:**
- testing
- Image card

Important: This page is not yet visible and routed to. But due to its size, we will merge parts of the page.

**What it looks like:**
image


MelanieTheLion:
Something went wrong with the files. It kept showing code that was already in main as 'changed/added'.

PR #247, opened 2026-04-30
Refactor Digital Specimen overview: Data cards

MelanieTheLion:
**In this PR:**
- Created first set up with the Digital Specimen cards for the overview page with their respective data
- Cards will not be shown if there is no data for that specific subject
- The hero with the specific info on the digital specimen

**Not in this PR:**
- testing
- Image card

Important: This page is not yet visible and routed to. But due to its size, we will merge parts of the page.

**What it looks like:**
image


src/utils/DataMappers/schemas/DigitalSpecimenSchema.ts
```diff
@@ -41,22 +41,23 @@ const DIGITAL_SPECIMEN_SCHEMA_MAP: Record<string, Record<string, FieldConfig>> =
                 const htmlLabel = acceptedIdentification?.["ods:scientificNameHTMLLabel"];
                 const fallbackLabel = acceptedIdentification?.["dwc:scientificName"];
                 
-                return {
```
MelanieTheLion, line 44: Removed because I broke other things.

src/components/Cards/DigitalSpecimenCard/DigitalSpecimenCard.scss
```diff
@@ -0,0 +1,33 @@
...
+    }
+    .ds-card-georeference {
+        width: 100%;
+        height: var(--general-card-width);
+        margin-block-end: var(--spacing-sm);
+    }
+    .ds-card-citation {
+        max-height: var(--citation-content-spacing);
+        width: 100%;
+        background-color: #F2F3F8;
```
JontySponselee-Naturalis: Does this need a css var?

src/components/Hero/Hero.scss
```diff
@@ -5,6 +5,12 @@ header {
         display: flex;
         justify-content: space-between;
         margin-block-end: var(--spacing-l);
+
+        div {
```
JontySponselee-Naturalis: using html tags, especially `div` could lead to hard to find css bugs. Is this css specificity needed? Can't you use a class or attribute selector?

src/components/Hero/Hero.tsx
```diff
@@ -90,14 +98,21 @@ export const Hero = ( { title, description, badge, navigateTo, showShareButton,
...
+                    if (!content) return null;
+                    return (
+                        <Badge color={color} variant={type} key={content}>{content}</Badge>
+                    )
+                })}
+            </div>
             <div id="hero-title">
-                <h1>{title}</h1>
+                {isHtml ? (
+                    <h1 dangerouslySetInnerHTML={{ __html: title }} />
```
JontySponselee-Naturalis: Where does `title` come from, is it generated by users or external systems? because this can introduce injections vulnerabilities. If it is, then you need to sanitize it

src/components/LabelValuePair/LabelValuePair.tsx
```diff
@@ -0,0 +1,82 @@
...
+        value: string;
+        isHtml: boolean;
+        type: string;
+		hidden: boolean;
+    }
+}
+
+/**
+ * Renders a single row of data.
+ * If the value is missing/null, it returns null to hide the row.
```
JontySponselee-Naturalis, line 29: When is it missing, is this normal behavior, or should it be logged so you can investigate it?

src/components/LabelValuePair/LabelValuePair.tsx
```diff
@@ -0,0 +1,82 @@
...
+				return (
+					<button className="btn-as-link" onClick={() => copy(item?.value)}>
+						{item?.value.replace(RetrieveEnvVariable('DOI_URL'), '')}
+						<CopyIcon />
+					</button>
+				)
+			default:
+				return (
+					item?.isHtml ? (
+						<span dangerouslySetInnerHTML={{ __html: item.value }} />
```
JontySponselee-Naturalis: same question as before for the `title`

src/components/Hero/Hero.tsx
MelanieTheLion: It is not generated by users, it is either a hardcoded title from me or data that comes from our own API's. Do you recommend sanitizing it then and if so, how?

src/components/LabelValuePair/LabelValuePair.tsx
MelanieTheLion: No, sometimes the data is just not there. Like for instance taxonomic data. It sometimes has a family, a genus etc, but not a species name for instance. Then we just don't show anything. So we do not have to investigate it, and only show the data that is available :)

src/components/Cards/DigitalSpecimenCard/DigitalSpecimenCard.scss
MelanieTheLion: Yes!! Changed!

src/components/Hero/Hero.scss
MelanieTheLion: Yes, changed!!

src/components/Hero/Hero.tsx
JontySponselee-Naturalis: It is fine if it is your input. However, if it comes from your API then it depends where it originally  comes from, so in that case I would recommend to santize it. 

I have not done this myself, so I can't help much, but I would recommend to use a package since it is security related. But this package seems to be used a lot https://www.npmjs.com/package/dompurify

MelanieTheLion: Ok, I've updated it with the package you mentioned. I saw it in multiple places as well. So I think it should be good. :)

src/components/LabelValuePair/LabelValuePair.tsx
MelanieTheLion: Same answer as above, adjusted now!

src/components/Hero/Hero.tsx
```diff
@@ -1,6 +1,7 @@
 /* Import dependencies */
 import { useLayoutEffect, useRef, useState } from "react";
 import { format } from "date-fns";
+import DOMPurify from 'dompurify';
```
JontySponselee-Naturalis: It is fine to import it directly, an alternative is to use a wrapper. This is useful when more components are using it. So you can easily swap it later with another packages and have one point of entry

MelanieTheLion: Ok, added it as a util :)

JontySponselee-Naturalis (approved):

PR #249, opened 2026-05-15
Added form inputs + state management + validation + testing to the Create VC form

MelanieTheLion:
In this PR:
- Form inputs according to the design
- State management (Zustand) for the first time as state management for the form
- Native validation for the inputs on 'next'
- A11y check on the multistepform & views

Design: https://www.figma.com/design/CW7NHzRh3Eu5555lxqNrUr/DiSSCover-v2---UX-design?node-id=510-28775&t=ewDmcLHTtWN9R3jF-0

Not in this PR:
- Unit testing multistepform 
- Mobile version
- e2e testing, this will be done once the 'create' button will be enabled and the POST is added

What it looks like:
Screenshot from 2026-05-21 10-23-31

Screenshot from 2026-05-21 10-24-08

Screenshot from 2026-05-21 10-23-05


src/components/MultiStepForm/MultiStepForm.tsx
```diff
@@ -26,12 +26,22 @@ interface Props {
 const MultiStepForm = ({ steps, handleCancel, handleSubmit }: Props) => {
     /* Base variables */
     const [currentStep, setCurrentStep] = useState(0);
+    const formRef = useRef<HTMLFormElement>(null);
+    const [wasValidated, setWasValidated] = useState(false);
```
JontySponselee-Naturalis, line 33: I do wonder, for this  whole multiStepform feature that it would make more sense to use react-hook-for package instead, because you are now building validation logic into it which this package fully cover.

src/store/useVirtualCollectionStore.ts
```diff
@@ -0,0 +1,36 @@
...
+    specimenRawList: string
+}
+
+interface VirtualCollectionState {
+    formData: CreateVirtualCollectionData,
+    updateField: (field: any, value: any) => void;
+    resetFormData: () => void
+}
+
+export const useVirtualCollectionStore = create<VirtualCollectionState>((set) => ({
```
JontySponselee-Naturalis: I am not sure that zustand is the right choice for this feature. Zustand is a global store, which means this data will be kept in memory even if you visit other pages, is this desired? I would expect that this should be isolated to where the multistep form is used. I am not sure if zustand does support this isolation. But I know that Jotai has this feature, I have used that as well for SAIL's multi-step-forms

src/components/MultiStepForm/views/AboutView.tsx
```diff
@@ -1,11 +1,77 @@
...
     return (
-        <div>
-            <h2>Temporary about view</h2>
+        <div id="about-view" className="form-view-container">
+            <h2>About this collection</h2>
+            <p>Enter title, description and the type of your virtual collection. In the next step, you can add the list of specimens to include in this collection.</p>
+
+            {/* TITLE FIELD */}
+            <div className="input-group">
+                <label htmlFor="form-title">Title</label>
```
JontySponselee-Naturalis: FYI, using `htmlFor` is fine, but you can also wrap `label` around `input` if you like that, then you also don't need the `id` in the `input`

src/components/MultiStepForm/views/AboutView.tsx
```diff
@@ -1,11 +1,77 @@
...
+                    id="form-title"
+                    name="title"
+                    value={title} 
+                    onChange={(e) => updateField('title', e.target.value)}
+                    required
+                    maxLength={128}
+                    aria-describedby="title-error"
+                    placeholder="Short and concise. Max 128 chars."
+                />
+                <div className="form-error" id="title-error" aria-live="polite">
```
JontySponselee-Naturalis, line 50: `aria-live` cool feature

src/components/MultiStepForm/views/AboutView.tsx
```diff
@@ -1,11 +1,77 @@
...
-        <div>
-            <h2>Temporary about view</h2>
+        <div id="about-view" className="form-view-container">
+            <h2>About this collection</h2>
+            <p>Enter title, description and the type of your virtual collection. In the next step, you can add the list of specimens to include in this collection.</p>
+
+            {/* TITLE FIELD */}
+            <div className="input-group">
+                <label htmlFor="form-title">Title</label>
+                <input
```
JontySponselee-Naturalis, line 37: Since you are adding arias, this one is also nice to have

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-invalid

src/components/MultiStepForm/views/AboutView.tsx
```diff
@@ -1,11 +1,77 @@
...
+                    <p>Please describe your virtual collection.</p>
+                </div>
+            </div>
+
+            <label htmlFor="form-type">Type</label>
+            <RadioCards.Root id="form-type" name="type">
+                <RadioCards.Item 
+                    value="Reference Collection" 
+                    checked
+                    onChange={(e) => updateField('type', (e.target as HTMLInputElement).value)}
```
JontySponselee-Naturalis: what type is `e`, kinda weird that you need to `as` this for radix?

if it has  `any`, then you can probably do something like `(e: React.ChangeEvent<HTMLInputElement>) => updateField('type', e.target.value) )`

src/components/MultiStepForm/views/ConfirmView.tsx
```diff
@@ -1,11 +1,109 @@
...
+const ConfirmView = ({ onEditAbout, onEditSpecimens }: ConfirmViewProps) => {
+    /* Base variables from the Virtual Collection store */
+    const title = useVirtualCollectionStore((state) => state.formData.title);
+    const description = useVirtualCollectionStore((state) => state.formData.description);
+    const type = useVirtualCollectionStore((state) => state.formData.type);
+    const specimen = useVirtualCollectionStore((state) => state.formData.specimen);
+
+    /* UI Toggle States */
+    const [showDescription, setShowDescription] = useState(false);
+    const [showAllSpecimen, setShowAllSpecimen] = useState(false); // <-- Restored state!
```
JontySponselee-Naturalis: `// <-- Restored state!` is a vague comment to me, why is this state special what do you mean with it?

src/components/MultiStepForm/views/ConfirmView.tsx
```diff
@@ -1,11 +1,109 @@
...
+    const [isDescriptionClamped, setIsDescriptionClamped] = useState(false);
+    
+    const descriptionRef = useRef<HTMLParagraphElement>(null);
+
+    /* Description line-clamp check */
+    useLayoutEffect(() => {
```
JontySponselee-Naturalis, line 31: I think this can become a nice utility function/component if you have more  text fields that requires this

src/components/MultiStepForm/views/SpecimensView.tsx
```diff
@@ -1,11 +1,60 @@
...
+    const updateField = useVirtualCollectionStore((state) => state.updateField);
+
+    /* Function to update the specimenRawList field in the store on input change */
+    const handleOnChangeTextarea = (e: ChangeEvent<HTMLTextAreaElement>) => {
+        updateField('specimenRawList', e.target.value);
+    }
+
+    /* Function to transform string to array of strings, clean it and check for duplicate items */
+    const handleOnBlurTextarea = () => {
+        const rawArray = specimenRawList.split(/[\n,;]+/);
```
JontySponselee-Naturalis: regexes always feel like magic, this is small and fairly understandable, but a comment would help more

src/components/MultiStepForm/views/SpecimensView.tsx
```diff
@@ -1,11 +1,60 @@
...
+        const cleanArray = rawArray
+            .map((item: string) => item.trim())
```
JontySponselee-Naturalis: minor thing, if a variable (this case `rawArray`) is well typed, then methods will correctly infer it, so it is a bit redundant to type `item: string` explicitly. There is something to say that being explicit is always good, so I think preference plays more of a role here

src/components/MultiStepForm/views/Views.scss
```diff
@@ -0,0 +1,158 @@
...
+
+    label {
+        display: block;
+        font-size: var(--sm-font-size);
+        font-weight: var(--medium-font-weight);
+        padding-block-end: var(--spacing-sm);
+        padding-block-start: var(--spacing-xl);
+    }
+
+    input, textarea {
```
JontySponselee-Naturalis: Won't you be using these HTML tags outside of the `.form-view-container`, shouldn't these styles be set in a Input component or globally. Or are these styles really coupled to this `.form-view-container`. 

And isn't the spacing tied to your design system (e.g. using ` var(--spacing-)` ?)

src/components/MultiStepForm/views/Views.scss
```diff
@@ -0,0 +1,158 @@
...
+        border-radius: 4px;
+        padding: 8px;
+        font-size: 14px;
+        border: 1px solid var(--gray-a4);
+        resize: none;
+        margin-block-end: 8px;
+    }
+
+    input:focus, textarea:focus {
+        outline: none;
```
JontySponselee-Naturalis: outline shouldn't be disabled, this is an essential a11y feature

src/components/MultiStepForm/views/Views.scss
```diff
@@ -0,0 +1,158 @@
+.form-view-container {
```
JontySponselee-Naturalis, line 1: Not all css properties uses the tied system design variables like fonts, padding etc

src/components/MultiStepForm/views/Views.scss
```diff
@@ -0,0 +1,158 @@
...
+
+/* Confirm view overrides */
+#confirm-view {
+    @extend .form-view-container;
+
+    .form-review-container {
+        padding: var(--spacing-sm);
+        background-color: var(--gray-a2);
+        margin-block-start: var(--spacing-xl);
+        border-radius: 4px;
```
JontySponselee-Naturalis: Is see a lot of `border-radius: 4px;` a css var is probably useful

src/components/MultiStepForm/views/Views.scss
```diff
@@ -0,0 +1,158 @@
...
+
+        .form-review-header {
+            display: flex;
+            justify-content: space-between;
+            align-items: center;
+        }
+        .clamped-text {
+            display: -webkit-box;
+            -webkit-box-orient: vertical;
+            -webkit-line-clamp: 3;
```
JontySponselee-Naturalis, line 175: best to avoid provider specific styling, if it is needed, then it is better to be handled by your bundler

src/store/useVirtualCollectionStore.ts
```diff
@@ -0,0 +1,36 @@
...
+    title: string,
+    description: string,
+    type: string,
+    specimen: string[],
+    specimenRawList: string
+}
+
+interface VirtualCollectionState {
+    formData: CreateVirtualCollectionData,
+    updateField: (field: any, value: any) => void;
```
JontySponselee-Naturalis: I think these `any`s should probably be generics based on the state interface probably. If it is tied to `CreateVirtualCollectionData`, then you can do `<K extends keyof CreateVirtualCollectionData>(field: K, value: CreateVirtualCollectionData[K]) => void;`

src/styles/typography.scss
```diff
@@ -49,3 +49,17 @@ a.navigation-link {
...
+    border: none;
+    padding: 0;
+    margin: 0;
+    background: none;
+    font-size: 12px;
+    color: var(--indigo-11);
+    cursor: pointer;
+    text-decoration: underline;
+    font-weight: var(--medium-font-weight);
+    letter-spacing: 0.04px;
```
JontySponselee-Naturalis: i was not aware that this specificity was possible lol

src/components/MultiStepForm/MultiStepForm.tsx
MelanieTheLion: I've thought about whether I should use a library or not. But I wanted to do it as native as I could. So for now this is a conscious choice :)

src/store/useVirtualCollectionStore.ts
MelanieTheLion: Zustand is removed! It was way too overpowered here, and didn't make sense to use the global store. I left out Jotai as well, because I could solve it with a local useState perfectly!

src/components/MultiStepForm/views/AboutView.tsx
MelanieTheLion: So I get this, and it might make sense. But I decided to keep the htmlFor to make my styling easier. I tried wrapping a label around the input, and I got into styling troubles and it didn't eliminate the need for the `id`

src/components/MultiStepForm/views/AboutView.tsx
MelanieTheLion: Added!

src/components/MultiStepForm/views/AboutView.tsx
MelanieTheLion: Removed!

src/components/MultiStepForm/views/ConfirmView.tsx
MelanieTheLion: Removed!

src/components/MultiStepForm/views/ConfirmView.tsx
MelanieTheLion: For now, I left it in the component. But will keep an eye on reusability and utilities!

src/components/MultiStepForm/views/SpecimensView.tsx
MelanieTheLion: Added comment!

src/components/MultiStepForm/views/SpecimensView.tsx
MelanieTheLion: Removed :dancers:

src/components/MultiStepForm/views/Views.scss
MelanieTheLion: Set globally now in my scss. BUT because I have Radix, Bootstrap and my own styling, I ran into 'order' issues with this. The border-radius and border attributes were being overwritten by Bootstrap. So instead of using `!important`, I have added a class to specifically target inputs and textareas as long as I still have Bootstrap.

src/components/MultiStepForm/views/Views.scss
MelanieTheLion: OOPS. Adjusted!

src/components/MultiStepForm/views/Views.scss
MelanieTheLion: All is tied to the design system now!

src/components/MultiStepForm/views/Views.scss
MelanieTheLion: css var is added!

src/components/MultiStepForm/views/Views.scss
MelanieTheLion: -web-kit is accepted by all browsers and [the way to set the line-clamp](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/line-clamp). line-clamp as a css attribute did not work at all.

src/store/useVirtualCollectionStore.ts
MelanieTheLion: Removed!

src/styles/typography.scss
MelanieTheLion: Haha removed, this makes NO sense to me as well

src/components/MultiStepForm/views/AboutView.tsx
```diff
@@ -1,11 +1,106 @@
...
+      ];
+
     return (
-        <div>
-            <h2>Temporary about view</h2>
+        <div id="about-view" className="form-view-container">
+            <h2>About this collection</h2>
+            <p>Enter title, description and the type of your virtual collection. In the next step, you can add the list of specimens to include in this collection.</p>
+
+            <div className="input-group">
```
MelanieTheLion, line 35: Didn't add fieldset for these two inputs, very consciously. But added fieldset for the radio button group

src/components/MultiStepForm/views/SpecimensView.tsx
```diff
@@ -1,11 +1,73 @@
...
+            <h2>Link specimen to collection</h2>
+            <ul>
+                <li>Paste your list of specimen identifiers or DOIs below</li>
+                <li>The identifiers / DOIs must be <b>globally unique</b> and <b>available in DiSSCo.</b></li>
+                <li>Separate items by line breaks, commas, or semicolons.</li>
+            </ul>
+
+            {/* SPECIMEN FIELD */}
+            <fieldset className="input-group">
+                <label htmlFor="form-specimen">Specimen DOIs</label>
```
ba55ie: A `fieldset` requires a `legend` or heading element. Since you're just wrapping one field, it's probably better to remove it here.

You could also use a role="group" instead of the class name for styling and semantics.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/group_role

src/styles/typography.scss
```diff
@@ -49,3 +49,29 @@ a.navigation-link {
...
+    color: var(--indigo-11);
+    cursor: pointer;
+    text-decoration: underline;
+    font-weight: var(--medium-font-weight);
+    text-underline-offset: var(--spacing-xs);
+    text-decoration-color: var(--indigo-5);
+}
+
+/* Form controls */
+input, textarea {
```
ba55ie: I would recommend using the `:where()` and `:is()` pseudo-classes for code organization and managing specificity, especially because your setting the defaults here. And combined with css nesting (which can be done natively nowadays).

Oh, and form inputs aren't really typography, right? ;)

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:where

src/styles/typography.scss
```diff
@@ -49,3 +49,29 @@ a.navigation-link {
...
+    width: 100%;
+    padding: var(--spacing-sm);
+    font-size: var(--sm-font-size);
+    resize: none;
+    margin-block-end: var(--spacing-sm);
+}
+
+input:focus, textarea:focus {
```
ba55ie: `:focus-visible` is the modern pseudo-class for focus ring management.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/:focus-visible

src/components/MultiStepForm/views/SpecimensView.tsx
MelanieTheLion: Removed :)

src/styles/typography.scss
MelanieTheLion: Haha no, moved to general.scss ! And also added :is!

src/styles/typography.scss
MelanieTheLion: Changed!

src/components/MultiStepForm/views/SpecimensView.tsx
MelanieTheLion: Removed!

src/styles/typography.scss
MelanieTheLion: Changed into :is, and hahaha no they are not typography. Moved it to general.scss

src/styles/typography.scss
ba55ie: You can nest this within the `:is(input, textarea) {}` rule.

ba55ie (approved):
There's only some inconsistent css nesting, but for the rest it looks fine by me. :)

JontySponselee-Naturalis (approved):

PR #251, opened 2026-06-03
Removed about page from navigation and renamed data export to export data

MelanieTheLion:
Very small VC UI improvements


Eleni-Theodoridou (approved):
🍏

PR #252, opened 2026-06-08
Added the POST service for the Create Virtual Collection flow

MelanieTheLion:
In this PR:
- POST service via Axios
- useMutation hook
- Available 'create' button that now works :smile: 

When a user goes through the Create a VC form flow, a new VC will now be created if the information the form contains is correct!

### DIFF
--- a/src/hooks/useVirtualCollections.ts
+++ b/src/hooks/useVirtualCollections.ts
@@ -7,6 +7,9 @@ import { getSelectedVirtualCollection } from 'services/virtualCollectionService/
 import { getVirtualCollectionDetails } from 'services/virtualCollectionService/getVirtualCollectionDetails';
 import { postNewVirtualCollection } from 'services/virtualCollectionService/postNewVirtualCollection';
 
+/* Import types */
+import { NewVirtualCollectionRequest } from 'types/Types';
+
 /* Base constants */
 const staleTime = 1000 * 60 * 5; // How long until the time is stale
 const gcTime = 1000 * 60 * 10; // Cache time: How long to store it in the cache
@@ -68,7 +71,7 @@ export const useSelectedVirtualCollection = ({ identifier }: { identifier: strin
  */
 export const useCreateVirtualCollection = () => {
     return useMutation({
-        mutationFn: (requestBody: any) => {
+        mutationFn: (requestBody: NewVirtualCollectionRequest) => {
             return postNewVirtualCollection(requestBody);
         }
     })
--- a/src/hooks/useVirtualCollections.ts
+++ b/src/hooks/useVirtualCollections.ts
@@ -1,10 +1,11 @@
 /* Import dependencies */
-import { useQuery } from '@tanstack/react-query';
+import { useMutation, useQuery } from '@tanstack/react-query';
 
 /* Import services */
 import { getAllVirtualCollections } from 'services/virtualCollectionService/getAllVirtualCollections';
 import { getSelectedVirtualCollection } from 'services/virtualCollectionService/getSelectedVirtualCollection';
 import { getVirtualCollectionDetails } from 'services/virtualCollectionService/getVirtualCollectionDetails';
+import { postNewVirtualCollection } from 'services/virtualCollectionService/postNewVirtualCollection';
 
 /* Base constants */
 const staleTime = 1000 * 60 * 5; // How long until the time is stale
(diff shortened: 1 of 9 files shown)


src/hooks/useVirtualCollections.ts
```diff
@@ -59,4 +60,16 @@ export const useSelectedVirtualCollection = ({ identifier }: { identifier: strin
...
     });
+}
+
+/**
+ * Hook that calls the postNewVirtualCollection service that creates a new Virtual Collection
+ * @returns The response of the service
+ */
+export const useCreateVirtualCollection = () => {
+    return useMutation({
+        mutationFn: (requestBody: any) => {
```
MelanieTheLion: Specify requestBody

src/pages/CreateVirtualCollection/CreateVirtualCollection.tsx
```diff
@@ -1,9 +1,11 @@
 /* Import components */
+import { RetrieveEnvVariable } from 'app/Utilities';
```
MelanieTheLion, line 2: Add order

src/services/virtualCollectionService/postNewVirtualCollection.ts
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Service that creates a new Virtual Collection by POST through the apiClient
+ * @returns The result of the create, either successful or error
+ */
+export const postNewVirtualCollection = async (requestBody: NewVirtualCollectionRequest) => {
+    try {
+        /* Call service and wait for response */
+        const response = await apiClient.post(
+            '/virtual-collection/v1',
```
Ycreak, line 14: is it correct to hard code _v1_ here?

MelanieTheLion: Yeah, we only have v1's as part of our endpoint structure. It is not a choice the user can make. I suspect that we might one day migrate to a yet-to-create v2 and v3 endsoforth.

src/pages/CreateVirtualCollection/CreateVirtualCollection.tsx
```diff
@@ -42,8 +46,32 @@ const CreateVirtualCollection = () => {
...
+                        "ods:predicateType": "equals",
+                        "ods:predicateKey": "$['dcterms:identifier']",
+                        "ods:predicateValues": collectionData.specimens
+                    }
+                }
+            }
+        }
+        mutate(requestBody, {
+            onSuccess: (response) => {
+                navigate(`${'/virtual-collections/' + response.id.replace(RetrieveEnvVariable('HANDLE_URL'), '')}`);
```
ba55ie: Just use the [Template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) without string concatenation (`+`).

src/services/apiClient.ts
```diff
@@ -1,10 +1,29 @@
+import KeycloakService from 'app/Keycloak';
 import axios from 'axios';
 
 const api = axios.create({
   baseURL: `${globalThis.location.protocol}//${globalThis.location.hostname}${globalThis.location.port ? ':' + globalThis.location.port : ''}/api`,
```
ba55ie: You can replace this with [`globalThis.location.origin`](https://developer.mozilla.org/en-US/docs/Web/API/Location/origin).

src/services/virtualCollectionService/postNewVirtualCollection.ts
Ycreak: yes, but should you not put that somewhere as a variable: then you only have to change it once (: not a big problem though, just might be a nice to have :D

src/services/apiClient.ts
MelanieTheLion: omg, lovely. Changed!

src/pages/CreateVirtualCollection/CreateVirtualCollection.tsx
MelanieTheLion: Whaha, I confuse myself sometimes. Changed!

src/services/virtualCollectionService/postNewVirtualCollection.ts
MelanieTheLion: Maybe, let me first talk to Sam my API guy about what their idea of the v1 is. I'll write it down for a coming sprint, is that ok?

ba55ie-nbc: Wise choice. It's an agile practice to not add functionality early. And there are many opinions on api versioning; whether it's a good or bad practice. ;)

ba55ie (approved):
Looks fine to me 👍

PR #253, opened 2026-06-12
Cleared out trivyignore

MelanieTheLion:
Removed everything to see what I can clean up, what can be solved and what has to stay because I cannot remove it.


samleeflang (approved):
Top 🥇

PR #255, opened 2026-06-15
Created a mock test suite for most of my tests and actual api tests for the navigation suite

MelanieTheLion:
Improved E2E testing strategy:
For the bulk of the E2E tests I don't care what kind of response the API gives me. I am not testing my API's. I am testing my UI. So for most of my E2E tests, I am using a mocked response of the requested url. This way, when for instance the amount of Virtual Collections changes, I am not dependent on the content and my tests will not fail.

This way my tests will only fail if I break my own UI or functionality.

Whether or not my APIs give me the response I want is a different kind of test.

Note:
For the navigation E2E tests, I am actually using the actual API responses still, to keep a bit of dependency on APIs.

image


.github/workflows/.trivyignore
```diff
@@ -14,6 +14,9 @@ CVE-2026-33671
 # underscore is a fixed dependency of jsonpath and cannot be upgraded by us Jun-12-2026
 CVE-2026-27601
 
+# form-data is a dependency of two other packages and it cannot be automatically updated Jun-18-2026
```
JontySponselee-Naturalis, line 17: Sometimes it is possible to force with an explicit override, have you tried that? 

https://pnpm.io/settings#overrides

JontySponselee-Naturalis (approved):

.github/workflows/.trivyignore
MelanieTheLion: I am still on NPM with this application because I ran into issues I don't want to fix in old code. But I know that npm also allows overrides. My larger goal is to get rid of as many dependencies I can :) A further user story tackles my dependencies and pnpm migration, and I'll take the overrides with me then.

PR #256, opened 2026-06-30
Added e2e tests for accessibility and Create Virtual Collection flow without Keycloak mock

MelanieTheLion:
**In this PR:**
- Accessibility test for all 3 views of the multistep form with a fix for one issue
- E2E test for the Create Virtual Collection flow with filling in inputs and testing validation.

**What is not in this PR:**
An adequate mock of the roles in Keycloak. I have not been able to mock this properly, so I found a workaround (go to the /create url directly). I need more time to figure this out.

### DIFF
--- a/src/components/MultiStepForm/views/AboutView.tsx
+++ b/src/components/MultiStepForm/views/AboutView.tsx
@@ -17,16 +17,6 @@ const AboutView = ({data, onUpdate, wasValidated}: AboutViewProps) => {
     const isTitleInvalid = wasValidated && !data.title.trim();
     const isDescriptionInvalid = wasValidated && !data.description.trim();
 
-    const typeOptions = [
-        {
-          id: 'type-reference',
-          value: 'Reference Collection',
-          title: 'Reference Collection',
-          description: 'When publishing a gold-standard collection for others to use as an identification key.',
-          disabled: false,
-        }
-      ];
-
     return (
         <div id="about-view" className="form-view-container">
             <h2>About this collection</h2>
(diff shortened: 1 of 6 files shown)


src/components/MultiStepForm/views/AboutView.tsx
```diff
@@ -90,7 +90,7 @@ const AboutView = ({data, onUpdate, wasValidated}: AboutViewProps) => {
                         />
                         
                         <label htmlFor={option.id} className="radio-card-label">
-                            <h4 className="radio-card-title">{option.title}</h4>
+                            <p className="radio-card-title">{option.title}</p>
                             <p className="radio-card-description">{option.description}</p>
```
ba55ie-nbc: Only `inline/phrasing` content is allowed within a `<label>`, so a `<p>` is not.

tests-e2e/accessibility.spec.ts
```diff
@@ -71,4 +71,47 @@ test.describe('Accessibility', () => {
 		// And the results should be 0
 		expect(results.length).toBe(0);
     });
+});
+
+test.describe('Accessibility', () => {
+    test('Create a Virtual Collection form flow', async ({ page }) => {
+        await mockGetVirtualCollections(page);
+		await mockGetSelectedVirtualCollection(page);
+		await mockGetVirtualCollectionDetails(page);
```
ba55ie-nbc: Can you merge these `await` statements in a `await Promise.all()`? Or even better move them to a helper function, because you're repeating this a couple of times in your tests.

tests-e2e/test-utils.ts
```diff
@@ -37,5 +37,6 @@ export async function checkA11y(page: any, pageName: string, options?: any) {
 export const testUrls = {
```
ba55ie-nbc: I wouldn't add the urls to the `test-utils`. I think they should be defined in the test suites themselves.

Less coupling, easier to read and how often do they really change (KISS). 😉 

```
 test('About page components', async ({ page }) => {
        await page.goto('/about');

       // Test stuff...
 });
```

ba55ie-nbc (changes requested):
Really cool that you have automated `a11y` tests in your project! 👍

src/components/MultiStepForm/views/AboutView.tsx
MelanieTheLion: Ok, I did something completely different here. Only one radio button option is weird, and I did that because we _might one day_ get another option. 
I now changed this from a form control to simple html. There is no need yet to offer a radio button. The fieldset, legend and label looked weird there as well. The value is always static.

Let me know what you think here.

tests-e2e/accessibility.spec.ts
MelanieTheLion: Yes I can :) Beautiful suggestion. Did I do it how I should? If I understand correctly, the promises are fulfilled and that is enough. I don't need to do anything after that.

tests-e2e/test-utils.ts
MelanieTheLion: Fair :P Overengineered it a bit. Changed it!

ba55ie-nbc (approved):

tests-e2e/accessibility.spec.ts
ba55ie-nbc: Nice, I would just return the Promise and `await` this async function in your tests. Otherwise it's a bit strange that you aren't "awaiting" an async function.

```
export async function setUpMockData(page: any) {
    return Promise.all([/* your mocks */]);
}

// and in your tests
await setUpMockData(page);

MelanieTheLion: Adjusted! DANK

PR #257, opened 2026-06-30
Added active state to ui and fixed dropdown

MelanieTheLion:
In this PR:
- Added an active state for the buttons to let the user know where they are :)
- Fixed the dropdown, the a and button behaved differently and needed to become consistent.


Eleni-Theodoridou (approved):
🥇

PR #258, opened 2026-06-30
Made digital specimen cards more specific and according to their UI

MelanieTheLion:
**In this PR:**
The DigitalSpecimenCard was first a reusable component for all 5 dataCards. But I ran into issues that made the reusable component become way too complicated. So I decided to separate concerns and create separate components for each data card. More code and more components, but easier to maintain in the long run.

UI still needs tweaks, but I have outstanding questions I have to solve first.

This page is not officially live yet. Because it is so big, we are working in iterations.

[UX Design](https://www.figma.com/design/CW7NHzRh3Eu5555lxqNrUr/DiSSCover-v2---UX-design?node-id=519-22353&t=XOrP13jUmj5uFlgu-0)

**Not in this pr:**
- Image functionality
- A11y test
- Testing

**What it looks like now:**
image


src/components/Cards/Cards.scss
```diff
@@ -1,6 +1,27 @@
...
+        align-items: center;
+        padding-block-end: var(--spacing-sm);
+
+        h2 {
+            font-size: var(--md-font-size);
+            font-weight: var(--bold-font-weight);
+        }
+    }
+    .digital-specimen-card-content {
+        padding: var(--spacing-m) 0px;
```
MelanieTheLion: Add 0px to spacings, I use it a lot

src/components/Cards/Cards.scss
```diff
@@ -24,10 +45,9 @@
 
         }
     }
-    .ds-card-body {
-        display: grid;
-        grid-template-columns: var(--general-card-height) 1fr;
-        gap: var(--spacing-sm) var(--spacing-m);
-        align-items: start;
+    .card-divider {
+        height: 1px;
```
MelanieTheLion, line 49: Add 1px as well to odd spacings

src/components/Cards/Cards.scss
```diff
@@ -24,10 +45,9 @@
...
+        width: 100%;
+        background-color: var(--gray-5);
```
MelanieTheLion: Check this gray with the colors I have, I have a variable for it I think

src/components/Cards/CitationLicenseCard.tsx
```diff
@@ -0,0 +1,55 @@
+import { Button, Card } from '@radix-ui/themes';
+import './Cards.scss';
+import { LabelValuePair } from 'components/LabelValuePair/LabelValuePair';
+import { CopyIcon } from '@radix-ui/react-icons';
+
+interface Props {
+    fragment: any;
```
MelanieTheLion: No any pls

src/components/Cards/CitationLicenseCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+export const CitationLicenseCard = ({ fragment }: Props ) => {
```
MelanieTheLion, line 21: Also, this one is not done yet

src/components/Cards/CitationLicenseCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+    return (
+        <Card className="digital-specimen-card">
+            <div className="digital-specimen-card-header">
+                <h2>Citation and License</h2>
+                <Button variant="ghost">
+                    Copy citation
+                    <CopyIcon />
+                </Button>
+            </div>
+            <div className="ds-card-citation">
```
MelanieTheLion: Choose between ds and digital-specimen as class names

src/components/Cards/CitationLicenseCard.tsx
```diff
@@ -0,0 +1,55 @@
...
+                <p>{craftCitation(fragment)}</p>
+            </div>
+            <div className="digital-specimen-card-content">
+                <LabelValuePair item={fragment['license']}></LabelValuePair>
```
MelanieTheLion, line 63: License needs to be url

src/components/LabelValuePair/LabelValuePair.scss
```diff
@@ -3,13 +3,21 @@
...
 
     .verbatim {
         font-size: var(--xs-font-size);
         color: var(--dark-gray);
+        font-weight: var(--regular-font-weight);
+    }
+
+    .taxonomic-status {
+        a {
+            margin-inline-end: 8px;
```
MelanieTheLion: variable pls

src/components/LabelValuePair/LabelValuePair.tsx
```diff
@@ -21,14 +18,21 @@ type Props = {
         isHtml: boolean;
         type: string;
 		hidden: boolean;
-    }
+    },
+	extraItem?: {
```
MelanieTheLion: Is this the way to go here?

src/components/LabelValuePair/LabelValuePair.tsx
```diff
@@ -72,10 +80,7 @@ export const LabelValuePair = ({ item }: Props) => {
   
     return (
 		<div className="property-row-fragment">
-			{item.type === 'verbatim' ?
-				<span className="property-label verbatim">{item.label}</span>
-				: <span className="property-label">{item.label}</span>
-			}
+			<span className={"property-label " + (item.type === 'verbatim' ? "verbatim" : "")}>{item.label}</span>
```
MelanieTheLion: This can be differently formatted

src/utils/DataMappers/schemas/DigitalSpecimenSchema.ts
```diff
@@ -29,8 +29,12 @@ const DIGITAL_SPECIMEN_SCHEMA_MAP: Record<string, Record<string, FieldConfig>> =
             resolve: (ds) => ds["dwc:basisOfRecord"]
         },
         discipline: {
-            label: 'Discipline',
+            label: 'Topic Discipline',
             resolve: (ds) => ds["dwc:topicDiscipline"]
+        },
+        virtualCollection: {
+            label: 'In Virtual Collection(s)',
+            resolve: (_) => 'Unknown'
```
MelanieTheLion: Check with Martijn what to do here

src/components/Cards/Cards.scss
MelanieTheLion: Changed to `--medium-gray`

src/components/Cards/CitationLicenseCard.tsx
MelanieTheLion: Changed to very specific types :)

src/utils/DataMappers/schemas/DigitalSpecimenSchema.ts
MelanieTheLion: Could be removed

src/components/Cards/CitationLicenseCard.tsx
MelanieTheLion: Now it is :)

src/components/LabelValuePair/LabelValuePair.scss
MelanieTheLion: done

src/components/Cards/Cards.scss
```diff
@@ -0,0 +1,53 @@
...
+        background-color: var(--indigo-2);
+        padding: var(--spacing-m);
+        margin-block-end: var(--spacing-sm);
+
+        p {
+            font-size: var(--sm-font-size);
+
+            a {
+                color: inherit;
+                text-decoration: underline;
```
JontySponselee-Naturalis, line 43: shouldn't this actually be the default? It is probably removed because a is also styled as a button?

src/components/Cards/CitationLicenseCard.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { LabelValuePair } from 'components/LabelValuePair/LabelValuePair';
+import { CopyIcon } from '@radix-ui/react-icons';
+
+/* Import styles */
+import './Cards.scss';
+
+type FragmentType = 'verbatim' | 'copy' | 'default' | 'url';
+
+interface Props {
+    fragment: {
```
JontySponselee-Naturalis, line 12: if `isHtml` is always `false` why even include it?. This is the case for multiple interfaces across files

src/components/LabelValuePair/LabelValuePair.tsx
```diff
@@ -8,78 +8,106 @@ import { Badge } from '@radix-ui/themes';
...
- * If the value is missing/null, it returns null to hide the row.
- */
-export const LabelValuePair = ({ item }: Props) => {
-	/* Base variables */
-	const { copy } = useClipboard();
+export const LabelValuePair = ({ item, extraItem }: Props) => {
+    const { copy } = useClipboard();
 
-	if (!item.value || item.hidden) return null;
+    if (item.value === undefined || item.value === null || item.value === '' || item.hidden) {
```
JontySponselee-Naturalis, line 54: `item.value === undefined || item.value === null || item.value === ''` 

This seems like a problem that should be solved when fetching this kind of data. As in, that you should transform the data before using it in the render components.  There shouldn't be 3 ways to indicate that a value is empty. There will probably be more times in the future that you need to check if it is empty, then there is a change that you will forget 1 or more of the empty options

src/components/Cards/CitationLicenseCard.tsx
MelanieTheLion: In IdentificationCard and GeoreferenceCard isHtml can be true (or false) or should be true, so I have included it.

src/components/Cards/IdentificationCard.tsx
```diff
@@ -0,0 +1,77 @@
+/* Import components */
+import { Button, Card } from '@radix-ui/themes';
+import { LabelValuePair } from 'components/LabelValuePair/LabelValuePair';
+import { Pencil2Icon } from '@radix-ui/react-icons';
+
+/* Import styles */
+import './Cards.scss';
+
+type FragmentType = 'verbatim' | 'copy' | 'default' | 'url';
+type isHtmlFragmentType = 'verbatim' | 'copy' | 'default';
```
ba55ie-nbc, line 10: You can use [Type Utilities](https://www.typescriptlang.org/docs/handbook/utility-types.html) and/or extending and merging types for creating new types from others.

```ts
type isHtmlFragmentType = Exclude<FragmentType, 'url'>;
```

src/components/Cards/IdentificationCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+interface Props {
+    fragment: {
+        scientificName: { label: string, value: string, isHtml: true, type: isHtmlFragmentType, hidden: boolean },
```
ba55ie-nbc, line 14: There's a lot of repetition going on with these types (in multiple files). There's the [rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)): When you're repeating the same (kind of) code 3 times, it's time for an abstraction.

src/components/Cards/IdentificationCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <Card className="digital-specimen-card">
+            <div className="digital-specimen-card-header">
+                <h2>Identification</h2>
+                <Button variant="ghost">
+                    Annotate
+                    <Pencil2Icon />
+                </Button>
+            </div>
+            <div className="digital-specimen-card-content">
+                <LabelValuePair item={fragment['scientificName']} ></LabelValuePair>
```
ba55ie-nbc, line 47: This looks like a [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl) (`<dl>`) to me. Also when taking the design into consideration.

src/components/LabelValuePair/LabelValuePair.tsx
ba55ie-nbc, line 1: And here we have the term `<dt>` and description `<dd>` elements for the parent `<dl>`.

You are allowed to wrap them in a `<div>` to make rows: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl#wrapping_name-value_groups_in_div_elements

src/components/Cards/IdentificationCard.tsx
ba55ie-nbc: Maybe that's something for a larger refactor later?

src/components/Cards/Cards.scss
MelanieTheLion: No, color: inherit means the color is black, instead of the normal link color --indigo-11, and our links have `text-decoration: none`, but this citation has `text-decoration: underline`. It breaks with our normal link styling, because it is styled as a citation box now :)

MelanieTheLion:
Over engineered it, should not have done that.

PR #261, opened 2026-07-17
Digital Specimen image functionality

MelanieTheLion:
**In this PR:**
- Possibility to show images, and show all available images in a list.
- When you click on an image, it is shown as the biggest image, and if you click on the biggest image, you are routed to the digital-media page

**Not in this PR:**
- Testing
- Accessibility

**What it looks like:**
image

### DIFF
--- a/src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
+++ b/src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
@@ -1,0 +1,89 @@
+import { RetrieveEnvVariable } from "app/Utilities";
+import { useEffect, useState } from "react";
+import { getJpegFromIIIFImages } from 'services/digitalMediaService/getJpegFromIIIFImages';
+
+/* Import styling */
+import './DigitalMediaCard.scss';
+
+interface Props {
+    specimen: any;
+}
+
+interface ImageFormat {
+    img: string;
+    id: string;
+}
+
+export const ImageCard = ({ specimen }: Props ) => {
+    /* Base variables */
+    const allImages = specimen?.digitalMedia?.map((item: any) => item['digitalMediaObject']) || [];
+    const [correctImageFormats, setCorrectImageFormats] = useState<ImageFormat[]>([]);
+    const [mainImage, setMainImage] = useState<string | null>(null);
+    const activeImageObject = correctImageFormats.find(item => item.img === mainImage);
+    const activeId = activeImageObject?.id || specimen?.digitalMedia?.[0]?.['digitalMediaObject']?.["@id"] || '';
+    const cleanDoiPath = activeId.replace(RetrieveEnvVariable('DOI_URL'), '');
+
+    /**
+     * This useEffect hook retrieves the jpg/jpeg url from IIIF images upon loading
+     * @returns Either jpg/jpeg id and img or null if image is not jpeg/jpg or iiif (json)
+     */
+    useEffect(() => {
+        (async () => {
+            if (!allImages.length) return;
+
+            const promises = allImages.map(async (img: any) => {
+                const format = img["dcterms:format"]?.toLowerCase();
+                const isJson = format === 'application/json';
+                const isJpeg = format === 'image/jpeg' || format === 'image/jpg';
+                
+                if (isJpeg) return { img: img["ac:accessURI"], id: img["@id"] };
+                if (isJson) return { img: await getJpegFromIIIFImages(img), id: img["@id"] };
+                return null;
+            });
+    
+            const results = await Promise.all(promises);
+            const filteredResults = results.filter((item): item is ImageFormat => item !== null);
+            
+            setCorrectImageFormats(filteredResults);
+
+            if (filteredResults.length > 0) {
+                setMainImage(filteredResults[0].img);
+            }
+        })();
+    }, [specimen]);
+
+    return (
+        <div className="digital-media-container">
+            <div className="digital-media-card">
+                {mainImage ? (
+                    <a href={`/dm/${cleanDoiPath}`}>
+                        <div 
+                            style={{ backgroundImage: `url(${mainImage})` }} 
+                            className="digital-media-card-image"
+                        ></div>
+                    </a>
+                ) : (
+                    <div className="digital-media-card-image placeholder">Loading...</div>
+                )}
+            </div>
+            
+            {correctImageFormats.length > 0 && (
+                <div className="digital-media-list">
+                    {correctImageFormats.map((image) => (
+                        <button
+                            key={image.id}
+                            onClick={() => setMainImage(image.img)}
+                            className="digital-media-thumb-btn"
+                        >
+                            <img
+                                src={image.img} 
+                                alt={`${image.id}`}
+                                className={image.img === mainImage ? 'active' : ''}
+                            />
+                        </button>
+                    ))}
+                </div>
+            )}
+        </div>
+    );
+};
--- a/src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
+++ b/src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
@@ -33,11 +33,9 @@ export const ImageCard = ({ specimen }: Props ) => {
 
             const promises = allImages.map(async (img: any) => {
                 const format = img["dcterms:format"]?.toLowerCase();
-                const isJson = format === 'application/json';
-                const isJpeg = format === 'image/jpeg' || format === 'image/jpg';
                 
-                if (isJpeg) return { img: img["ac:accessURI"], id: img["@id"] };
-                if (isJson) return { img: await getJpegFromIIIFImages(img), id: img["@id"] };
+                if (format === 'image/jpeg' || format === 'image/jpg') return { img: img["ac:accessURI"], id: img["@id"] };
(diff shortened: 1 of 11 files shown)


src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+    useEffect(() => {
+        (async () => {
+            if (!allImages.length) return;
+
+            const promises = allImages.map(async (img: any) => {
+                const format = img["dcterms:format"]?.toLowerCase();
+                const isJson = format === 'application/json';
+                const isJpeg = format === 'image/jpeg' || format === 'image/jpg';
+                
+                if (isJpeg) return { img: img["ac:accessURI"], id: img["@id"] };
```
southeo: maybe this is a matter of taste, but I'd just put the conditions here instead of creating new variables. isJson and isJpeg are also very similar variable names; at first i thought they were the same.

src/pages/DigitalSpecimenOverview/DigitalSpecimenOverview.tsx
```diff
@@ -18,15 +19,25 @@ const DigitalSpecimenDetails = () => {
     const segments = url.pathname.split('/');
     const identifier = segments.slice(2).join("/");
     const { data: specimen, isLoading, isError } = useDigitalSpecimenComplete({ doi: identifier});
+    const hasImages = specimen?.digitalMedia?.length > 0;
 
     if (isLoading) return <main><p>Retrieving the Digital Specimen Details...</p></main>;
     if (!specimen) return <main><p>No data found</p></main>
     if (isError) return <main><p>Something went wrong with fetching the Digital Specimen. Please try again later.</p></main>;
 
     const actualData = specimen?.mappedData || [];
```
southeo, line 29: What is this?

src/pages/DigitalSpecimenOverview/DigitalSpecimenOverview.tsx
```diff
@@ -18,15 +19,25 @@ const DigitalSpecimenDetails = () => {
     const segments = url.pathname.split('/');
     const identifier = segments.slice(2).join("/");
     const { data: specimen, isLoading, isError } = useDigitalSpecimenComplete({ doi: identifier});
+    const hasImages = specimen?.digitalMedia?.length > 0;
```
southeo, line 22: you could also use specimen.isKnownToContainMedia here if you wanted.

src/services/digitalMediaService/getJpegFromIIIFImages.ts
```diff
@@ -0,0 +1,29 @@
...
+
+        // IIIF v2 path and IIIF v3 fallback
+        const service = data.sequences?.[0]?.canvases?.[0]?.images?.[0]?.resource?.service || data.items?.[0]?.items?.[0]?.items?.[0]?.body?.service?.[0];
+
+        if (!service) {
+            throw new Error("Could not find an image service in this manifest.");
+        }
+
+        // IIIF v2 uses '@id', IIIF v3 uses 'id'
+        const imageServiceId = service['@id'] || service.id;
```
southeo, line 20: so is this essentially: 

if service[@id] is not null : imageServiceId = service['@id']
else imageServiceId = service['id']

what if they're both null?

southeo (commented):
🐐

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+    id: string;
+}
+
+export const ImageCard = ({ specimen }: Props ) => {
+    /* Base variables */
+    const allImages = specimen?.digitalMedia?.map((item: any) => item['digitalMediaObject']) || [];
+    const [correctImageFormats, setCorrectImageFormats] = useState<ImageFormat[]>([]);
+    const [mainImage, setMainImage] = useState<string | null>(null);
+    const activeImageObject = correctImageFormats.find(item => item.img === mainImage);
+    const activeId = activeImageObject?.id || specimen?.digitalMedia?.[0]?.['digitalMediaObject']?.["@id"] || '';
```
Ycreak, line 23:
```
specimen?.digitalMedia?.[0]?.['digitalMediaObject']?.["@id"]
```

oh javascript, such beautiful way of doing things

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
```diff
@@ -0,0 +1,89 @@
...
+    const cleanDoiPath = activeId.replace(RetrieveEnvVariable('DOI_URL'), '');
+
+    /**
+     * This useEffect hook retrieves the jpg/jpeg url from IIIF images upon loading
+     * @returns Either jpg/jpeg id and img or null if image is not jpeg/jpg or iiif (json)
+     */
+    useEffect(() => {
```
JontySponselee-Naturalis, line 30: If you are on react 19, you could also use the `use()` hook. Which will make the component in a "suspend" state while the promise is getting resolved. This will omit the first "flicker" render

https://react.dev/reference/react/use

src/services/digitalMediaService/getJpegFromIIIFImages.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export async function getJpegFromIIIFImages(image: any) {
+    try {
+        const response = await fetch(image["ac:accessURI"]);
+        if (!response.ok) throw new Error(`HTTP error! Status: ${response.status}`);
+
+        const data = await response.json();
+
+        // IIIF v2 path and IIIF v3 fallback
+        const service = data.sequences?.[0]?.canvases?.[0]?.images?.[0]?.resource?.service || data.items?.[0]?.items?.[0]?.items?.[0]?.body?.service?.[0];
```
Ycreak, line 13: what source are you reading here with so many question marks? shouldnt this be put in an interface?

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
Ycreak: same as my remark further down below, but what source are you reading? shouldnt specimen be an interface?

src/services/digitalMediaService/getJpegFromIIIFImages.ts
JontySponselee-Naturalis, line 20: Just to be sure, any of these are **always** filled in? (same remark as Sue, what if both are `null`)

src/pages/DigitalSpecimenOverview/DigitalSpecimenOverview.tsx
MelanieTheLion: Then it looks like this: specimen?.data?.attributes?.digitalSpecimen?.['ods:isKnownToContainMedia'] so I am keeping what I have now :dancers:

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
MelanieTheLion: We get data from all kinds of data sources. Although specimen is pretty well defined with the same keys in the object, because we create it like that, in this case digitalMedia could be an empty [] if there is no media. If so, the entire variable would through an error.

Ideally yes, it should be an interface. I think I should also look into this, but if it is ok with you I'd like to pick that up in a separate story.

src/services/digitalMediaService/getJpegFromIIIFImages.ts
MelanieTheLion: So this source is starting with a manifest.json, that points to a location for the IIIF images that I then have to retrieve in an impossible way. I have no control over de source, external, I could look into an interface, but again, then I'd like to do that in a separate user story. Because it might result in a lot of extra code.

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
MelanieTheLion: Good idea! I am not on React 19 yet and am waiting to upgrade until the bulk of all my old code is finally gone, to prevent me updated old code. I'd added your comment to my 'upgrade to react 19' user story though :)

src/services/digitalMediaService/getJpegFromIIIFImages.ts
MelanieTheLion: Very fair, I adjusted it to throw an error instead.

MelanieTheLion: Throwing an error now :)

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
MelanieTheLion: Adjusted! Don't want no unnecessary consts

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
Ycreak: Okay! That would be swell. Sometimes you just have to get data from an impossible source. An interface would be nice, but that is indeed a story for another time!

JontySponselee-Naturalis (approved):

PR #262, opened 2026-07-22
Initial set up of the VC images for no image

MelanieTheLion:
**In this PR:**
- A check on isKnownToContainMedia to show a 'no image' if the check comes back false
- Only shown on Mobile for now, because the gallery is not yet implemented on normal view

**What it looks like (on mobile still):**
image

### DIFF
--- a/src/components/Cards/VirtualCollectionCard/VirtualCollectionCard.tsx
+++ b/src/components/Cards/VirtualCollectionCard/VirtualCollectionCard.tsx
@@ -1,0 +1,40 @@
+import { Badge, Card } from "@radix-ui/themes";
+import { RetrieveEnvVariable } from "app/Utilities";
+import { GetSpecimenNameHTMLLabel } from "app/utilities/NomenclaturalUtilities";
+import { Link } from "react-router-dom";
+
+/* Import styling */
+import './VirtualCollectionCard.scss';
+
+interface Props {
+    collection: any,
+    type: "details" | "overview"
+}
+
+export const VirtualCollectionCard = ({ collection, type: viewType }: Props) => {
+    return (
+        <>
+            { viewType === "details" && 
+                <Card variant="surface" className="gallery-card" asChild>
+                    <Link to={`/ds/${collection.id.replace(RetrieveEnvVariable('DOI_URL'), '')}`}>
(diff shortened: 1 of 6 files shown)


src/components/Cards/VirtualCollectionCard/VirtualCollectionCard.tsx
```diff
@@ -0,0 +1,40 @@
...
+                <Card variant="surface" className="gallery-card" asChild>
+                    <Link to={`/ds/${collection.id.replace(RetrieveEnvVariable('DOI_URL'), '')}`}>
+                        { !collection.attributes['ods:isKnownToContainMedia'] &&
+                        <div className="vc-card-image-container">
+                            <span>No image</span>
+                        </div>
+                        }
+                        <div className="vc-card-content-container">
+                            { collection.attributes['ods:hasIdentifications'][0]['dwc:typeStatus'] && 
+                            <Badge color="sky" variant="solid">{collection.attributes['ods:hasIdentifications'][0]['dwc:typeStatus']}</Badge>
```
southeo, line 27: Do you need the `?` to prevent a possible null reference exception?

src/components/Cards/VirtualCollectionCard/VirtualCollectionCard.tsx
```diff
@@ -0,0 +1,40 @@
...
+                            }
+                            <p dangerouslySetInnerHTML={{__html: GetSpecimenNameHTMLLabel(collection.attributes)}}></p>
```
southeo, line 29: dangerously? 😨

southeo (approved):
🍾

src/components/Cards/VirtualCollectionCard/VirtualCollectionCard.tsx
MelanieTheLion: I added it :)

src/components/Cards/VirtualCollectionCard/VirtualCollectionCard.tsx
MelanieTheLion: Yes, the way within React to let React know that it is ok and I want to process the html I ingest here.

PR #264, opened 2026-07-28
Mobile & a11y check for the Digital Specimen Overview

MelanieTheLion:
**In this PR:**
- The mobile view for the digital specimen overview
- Trivyignore for react-router-dom. I added a story to my sprint to address the upgrade from 7 to 8.

**Design**: https://www.figma.com/design/CW7NHzRh3Eu5555lxqNrUr/DiSSCover-v2---UX-design?node-id=519-23133&t=Qq3Chw87fwYZRLRD-0

**Jira ticket**: https://naturalis.atlassian.net/browse/DD-3189

**What it looks like:**
image

image

### DIFF
--- a/.github/workflows/.trivyignore
+++ b/.github/workflows/.trivyignore
@@ -22,6 +22,9 @@ CVE-2026-12143
 # js-yaml is a dependency of json-schema-to-typescript and cannot be updated Jul-22-2026
 CVE-2026-59869
 
+# React-router-dom needs to be upgraded from 7 to 8 Jul-28-2926
+GHSA-qwww-vcr4-c8h2
+
 # Vulnerabilities in Image
 # libcrypto3 included in the image, cannot fix Jun-12-2026
 CVE-2026-45447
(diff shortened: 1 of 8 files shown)


.github/workflows/.trivyignore
```diff
@@ -22,6 +22,9 @@ CVE-2026-12143
 # js-yaml is a dependency of json-schema-to-typescript and cannot be updated Jul-22-2026
 CVE-2026-59869
 
+# React-router-dom needs to be upgraded from 7 to 8 Jul-28-2926
```
MelanieTheLion, line 25: I added a story to my sprint to pick this up ASAP: https://naturalis.atlassian.net/browse/DD-3208

src/components/Cards/DigitalMediaCard/DigitalMediaCard.scss
```diff
@@ -1,6 +1,13 @@
+@use "styles/mixins.scss" as *;
+
 .digital-media-container {
-    position: sticky;
-    top: 0;
+    position: initial;
```
JontySponselee-Naturalis, line 4: isn't this the default?

JontySponselee-Naturalis (approved):

src/components/Cards/DigitalMediaCard/DigitalMediaCard.scss
MelanieTheLion: You would think that, but it immediately goes to position sticky, even when it's not in the correct media query.

PR #265, opened 2026-07-28
Added LOADS of unit tests for multiple components

MelanieTheLion:
**In this PR:**
- Unit tests for several UI components

**Jira ticket:** https://naturalis.atlassian.net/browse/DD-2851

No E2E tests yet, because my digital specimen page is not yet published and thus I cannot get to it yet :)

### DIFF
--- a/src/components/Cards/DigitalMediaCard/DigitalMediaCard.test.tsx
+++ b/src/components/Cards/DigitalMediaCard/DigitalMediaCard.test.tsx
@@ -48,9 +48,12 @@ describe('DigitalMediaCard component', () => {
         vi.mocked(getJpegFromIIIFImages).mockResolvedValue('https://herbarium.bgbm.org/data/iiif/resolved_image.jpg');
     });
 
-    it('renders the loading placeholder initially before processing images', () => {
+    it('renders the loading placeholder initially before processing images', async () => {
         render(<DigitalMediaCard specimen={mockSpecimen} />);
+
         expect(screen.getByText('Loading...')).toBeInTheDocument();
+
+        await screen.findByAltText('https://doi.org/TEST/K05-DG7-0EM');
     });
 
     it('processes images, calls IIIF resolver for json, and renders thumbnails', async () => {
--- a/src/components/Cards/DigitalMediaCard/DigitalMediaCard.test.tsx
+++ b/src/components/Cards/DigitalMediaCard/DigitalMediaCard.test.tsx
@@ -1,0 +1,96 @@
+/* Import test dependencies */
+import { screen, fireEvent, render } from 'tests/test-utils';
+import { describe, it, expect, beforeEach, vi } from 'vitest';
+
+/* Import component & mocked modules */
+import { DigitalMediaCard } from './DigitalMediaCard';
+import { getJpegFromIIIFImages } from 'services/digitalMediaService/getJpegFromIIIFImages';
+
+/* Mock external services & utilities */
+vi.mock('services/digitalMediaService/getJpegFromIIIFImages', () => ({
+    getJpegFromIIIFImages: vi.fn(),
+}));
+
+vi.mock('app/Utilities', () => ({
+    RetrieveEnvVariable: vi.fn(() => 'https://doi.org/'),
+}));
+
+/* Mock data */
+const mockSpecimen = {
+    digitalMedia: [
+        {
+            digitalMediaObject: {
+                "@id": "https://doi.org/TEST/K05-DG7-0EM",
+                "ac:accessURI": "https://image.bgbm.org/images/internal/HerbarThumbs/B100630718_1700",
+                "dcterms:format": "image/jpeg",
+            }
+        },
+        {
+            digitalMediaObject: {
+                "@id": "https://doi.org/TEST/ZXE-EPN-B0V",
+                "ac:accessURI": "https://herbarium.bgbm.org/data/iiif/B100630718/manifest.json",
+                "dcterms:format": "application/json",
+            }
+        },
+        {
+            digitalMediaObject: {
+                "@id": "https://doi.org/TEST/IGNORE",
+                "ac:accessURI": "https://example.com/doc.pdf",
+                "dcterms:format": "application/pdf",
+            }
+        }
+    ]
+};
+
+describe('DigitalMediaCard component', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
+        vi.mocked(getJpegFromIIIFImages).mockResolvedValue('https://herbarium.bgbm.org/data/iiif/resolved_image.jpg');
+    });
+
+    it('renders the loading placeholder initially before processing images', () => {
+        render(<DigitalMediaCard specimen={mockSpecimen} />);
+        expect(screen.getByText('Loading...')).toBeInTheDocument();
+    });
+
+    it('processes images, calls IIIF resolver for json, and renders thumbnails', async () => {
+        render(<DigitalMediaCard specimen={mockSpecimen} />);
+
+        const firstThumb = await screen.findByAltText('https://doi.org/TEST/K05-DG7-0EM');
+        const secondThumb = await screen.findByAltText('https://doi.org/TEST/ZXE-EPN-B0V');
+
+        expect(firstThumb).toBeInTheDocument();
+        expect(secondThumb).toBeInTheDocument();
+
+        /* Check if IIIF service was called specifically for the json format */
+        expect(getJpegFromIIIFImages).toHaveBeenCalledTimes(1);
+        expect(getJpegFromIIIFImages).toHaveBeenCalledWith(mockSpecimen.digitalMedia[1].digitalMediaObject);
+
+        /* Verify that unsupported formats (e.g. application/pdf) are filtered out */
+        const queryUnsupported = screen.queryByAltText('https://doi.org/TEST/IGNORE');
+        expect(queryUnsupported).not.toBeInTheDocument();
+    });
+
+    it('updates main image background and DOI link path when clicking a thumbnail', async () => {
+        render(<DigitalMediaCard specimen={mockSpecimen} />);
+
+        /* Wait for first image thumbnail to load */
+        const secondThumb = await screen.findByAltText('https://doi.org/TEST/ZXE-EPN-B0V');
+        
+        /* Initial link should point to active DOI path of the first image */
+        const link = screen.getByRole('link');
+        expect(link).toHaveAttribute('href', '/dm/TEST/K05-DG7-0EM');
+
+        /* Click on the second thumbnail */
+        fireEvent.click(secondThumb);
+
+        /* Link should update to match the second image's DOI path */
+        expect(link).toHaveAttribute('href', '/dm/TEST/ZXE-EPN-B0V');
+    });
+
+    it('handles missing or empty specimen', async () => {
+        render(<DigitalMediaCard specimen={null} />);
+        expect(screen.getByText('Loading...')).toBeInTheDocument();
+        expect(screen.queryByRole('button')).not.toBeInTheDocument();
+    });
+});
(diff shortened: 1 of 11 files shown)


src/components/Cards/DigitalMediaCard/DigitalMediaCard.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+                "ac:accessURI": "https://example.com/doc.pdf",
+                "dcterms:format": "application/pdf",
+            }
+        }
+    ]
+};
+
+describe('DigitalMediaCard component', () => {
+    beforeEach(() => {
+        vi.clearAllMocks();
```
JontySponselee-Naturalis: if mocks always need to be cleared, you can add this as an option to `vite.config.ts`. This will make every test run in isolation

I have for example this in my config:

```
test: {
    clearMocks: true,
    mockReset: true,
    restoreMocks: true,
}
```

src/components/Cards/DigitalSpecimenCard/DigitalSpecimenCard.test.tsx
```diff
@@ -0,0 +1,213 @@
...
+        });
+    });
+
+    it('renders the header title and non-hidden fragment items', () => {
+        render(<DigitalSpecimenCard cardHeader="Specimen Overview" fragment={mockFragment} />);
+
+        expect(screen.getByRole('heading', { name: 'Specimen Overview' })).toBeInTheDocument();
+
+        /* LabelValuePair elements (5 visible items out of 6 in mockFragment) */
+        const renderedPairs = screen.getAllByTestId('label-value-pair');
```
JontySponselee-Naturalis: try to not use `test-id`.

I dont know the html structure, but `LabelValuePair` seems like a candidate for "description list"  which has a role you can test, or perhaps some other HTML semantic fits better then the expect below can check for the `dt`

src/components/Cards/DigitalSpecimenCard/DigitalSpecimenCard.test.tsx
```diff
@@ -0,0 +1,213 @@
...
+        it('renders map component with correct coordinates when georeference is true', () => {
+            render(
+                <DigitalSpecimenCard
+                    cardHeader="Test Card"
+                    fragment={mockFragment}
+                    georeference={true}
+                />
+            );
+
+            const map = screen.getByTestId('mock-map');
```
JontySponselee-Naturalis: `figure` can be used as HTML semantic for a map, which you can use as role to test

src/components/LabelValuePair/LabelValuePair.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+        it('renders null if hidden is set to true', () => {
+            const hiddenItem = {
+                label: 'Internal Code',
+                value: 'Secret 123',
+                isHtml: false,
+                type: 'string',
+                hidden: true,
+            };
+
+            // const { container } = render(<LabelValuePair item={hiddenItem} />);
```
JontySponselee-Naturalis: 💀

src/components/LabelValuePair/LabelValuePair.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+            const itemWithoutValue = {
+                label: 'Scientific Name',
+                value: '',
+                isHtml: false,
+                type: 'string',
+                hidden: false,
+            };
+
+            const { container } = render(<LabelValuePair item={itemWithoutValue} />);
+            expect(container.firstChild).toBeEmptyDOMElement();
```
JontySponselee-Naturalis, line 44: probably better to check with `queryByRole(...).not`

src/components/LabelValuePair/LabelValuePair.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+        it('renders standard text for default type when isHtml is false', () => {
+            const defaultItem = {
+                label: 'Country',
+                value: 'Netherlands',
+                isHtml: false,
+                type: 'string',
+                hidden: false,
+            };
+
+            render(<LabelValuePair item={defaultItem} />);
```
JontySponselee-Naturalis, line 71: Mentioned earlier but not sure if it is the same component, but this could probably use `dt` + `dd` HTML tags

src/components/LabelValuePair/LabelValuePair.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+                label: 'Description',
+                value: '<b>Panthera</b>',
+                isHtml: true,
+                type: 'string',
+                hidden: false,
+            };
+
+            render(<LabelValuePair item={htmlItem} />);
+
+            expect(sanitizeHtmlWrapper).toHaveBeenCalledWith('<b>Panthera</b>');
```
JontySponselee-Naturalis, line 88: It think it is better to check what this components renders instead of checking which internal function it uses

src/components/LabelValuePair/LabelValuePair.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+        });
+
+        it('renders Catalogue of Life link for "url" type', () => {
+            const urlItem = {
```
JontySponselee-Naturalis, line 93: it almost looks like you are building your own CMS renderer lol

src/components/LabelValuePair/LabelValuePair.test.tsx
```diff
@@ -0,0 +1,146 @@
...
+                label: 'Verbatim Locality',
+                value: 'Near the oak tree',
+                isHtml: false,
+                type: 'verbatim',
+                hidden: false,
+            };
+
+            render(<LabelValuePair item={verbatimItem} />);
+
+            expect(screen.getByText('Verbatim Locality')).toHaveClass('property-label verbatim');
```
JontySponselee-Naturalis: looking at classnames is as fragile as using test-ids better to test what it renders. If you need to test on visuals, better to have a look at https://main.vitest.dev/guide/browser/visual-regression-testing

src/components/Cards/DigitalSpecimenCard/DigitalSpecimenCard.test.tsx
MelanieTheLion: What is the role of the description list?[ MDN says that it does not have an implicit role](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl#:~:text=Implicit%20ARIA%20role,No%20corresponding%20role). I can add a role like list though. The component is a `<dl>`, so I had to change the HTML in my test anyway, because it misrepresented the truth.

src/components/LabelValuePair/LabelValuePair.test.tsx
MelanieTheLion: It has them, but not in this test suite, because here I don't have to mock this component. Do you expect it in this file?

src/components/LabelValuePair/LabelValuePair.test.tsx
MelanieTheLion: I agree, but if LabelValuePair is a `<dl>`, which role do I query?

src/components/LabelValuePair/LabelValuePair.test.tsx
MelanieTheLion: Partly. It is a utility function I test in another suite, so for this test I want to test whether I call it. But I did add a second assertion to test the outcome as well.

src/components/LabelValuePair/LabelValuePair.test.tsx
MelanieTheLion: It has been an adventure.

src/components/LabelValuePair/LabelValuePair.test.tsx
MelanieTheLion: I removed this verbatim test. I actually think that this test as it is right now is not as important for me. It goes beyond functionality of the component I need to test. I do like the visual regression testing advice, so I will look into that.
But not now :D

src/components/LabelValuePair/LabelValuePair.test.tsx
MelanieTheLion: WHAHA OEPS, removed

src/components/Cards/DigitalSpecimenCard/DigitalSpecimenCard.test.tsx
MelanieTheLion: Adjusted in the test

src/components/Cards/DigitalMediaCard/DigitalMediaCard.test.tsx
MelanieTheLion: Good one, added

src/components/Cards/DigitalSpecimenCard/DigitalSpecimenCard.test.tsx
JontySponselee-Naturalis: You can use the debug function `logRoles(document.body)` / `logRoles(aContainer)` to check which roles are available

https://testing-library.com/docs/dom-testing-library/api-debugging/#logroles

src/components/LabelValuePair/LabelValuePair.test.tsx
JontySponselee-Naturalis: see previous comment

src/components/LabelValuePair/LabelValuePair.test.tsx
JontySponselee-Naturalis: I did expect it in that file, actually I think it makes more sense to have `<dl />` as a component with props for keyValue pair instead of individual keyValue pairs. Another argument is because dl+dd+dt has  a specific structure according to the spec

src/components/Cards/DigitalSpecimenCard/DigitalSpecimenCard.test.tsx
MelanieTheLion: So a `dl` apparently has no role I can find with `logRoles(document.body)`. The dd and dt do have (description and term respectively). I am keeping the added 'role="list" here.

src/components/LabelValuePair/LabelValuePair.test.tsx
MelanieTheLion: Ok, I created a DescriptionList component for now, to at least have people use it like that :)

src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { UIProperty } from "types/dataMapperTypes";
+import { SpecimenField } from "types/digitalSpecimenTypes";
+
+interface Props {
+    fragment: SpecimenField[];
+}
+
+export const DescriptionList = ({ fragment }: Props) => {
+    return (
+        <dl className="ds-card-body">
```
JontySponselee-Naturalis, line 14: Je kan het eventueel nog meer uitbreiden met `aria-label` om je testen specifieker te maken en dat het ook een titel krijgt voor screen readers. Dit kan ook eventueel voor de `dd` en `dt` in `LabelValuePair`

src/types/digitalSpecimenTypes.ts
```diff
@@ -46,4 +46,11 @@ export const CARD_CONFIGS: Record<CardCategory, Partial<CardConfig> & { annotati
     },
     [CardCategory.CollectingEvent]: {},
     [CardCategory.CitationAndLicense]: { copy: true, citation: true }
-};
\ No newline at end of file
+};
+
+export type SpecimenField = {
+    label: string;
+    value: any;
```
JontySponselee-Naturalis, line 53: `unknown`  is type safer

JontySponselee-Naturalis (approved):

PR #266, opened 2026-07-30
Added Radix tabs (from Themes) to the Digital Specimen page

MelanieTheLion:
**In this PR:**
- Added Radix Tabs to the Digital Specimen page, but made it a reusable component still

**[Design for the tabs](https://www.figma.com/design/CW7NHzRh3Eu5555lxqNrUr/DiSSCover-v2---UX-design?node-id=73-11256&t=fNwpZ9kYoytDPhSL-0)** 

**Jira link:** https://naturalis.atlassian.net/browse/DD-2999

**What it looks like:**
image

image

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -24,15 +24,13 @@ export const DigitalSpecimenTabs = ({ defaultValue, tabs }: Props) => {
                 })}
             </Tabs.List>
 
-            <main>
-                { tabs.map((tab) => {
-                    return (
-                        <Tabs.Content key={'second-' + tab.value} value={tab.value}>
-                            {tab.component}
-                        </Tabs.Content>
-                    )
-                })}
-            </main>
+            { tabs.map((tab) => {
+                return (
+                    <Tabs.Content key={'second-' + tab.value} value={tab.value}>
+                        {tab.component}
+                    </Tabs.Content>
+                )
+            })}
         </Tabs.Root>
     )
 }
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -24,7 +24,7 @@ export const DigitalSpecimenTabs = ({ defaultValue, tabs }: Props) => {
                 })}
             </Tabs.List>
 
-            <div>
+            <main>
                 { tabs.map((tab) => {
                     return (
                         <Tabs.Content key={'second-' + tab.value} value={tab.value}>
@@ -32,7 +32,7 @@ export const DigitalSpecimenTabs = ({ defaultValue, tabs }: Props) => {
                         </Tabs.Content>
                     )
                 })}
-            </div>
+            </main>
         </Tabs.Root>
     )
 }
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,0 +1,38 @@
+import { Tabs } from "@radix-ui/themes";
+import { ReactNode } from "react";
+
+/* Import styling */
+import "./Tabs.scss";
+
+interface Props {
+    defaultValue: string,
+    tabs: {
+        value: string,
+        title: string,
+        component: ReactNode,
+    }[]
+}
+
+export const DigitalSpecimenTabs = ({ defaultValue, tabs }: Props) => {
+    return (
+        <Tabs.Root defaultValue={defaultValue}>
+            <Tabs.List>
+                { tabs.map((tab) => {
+                    return (
+                        <Tabs.Trigger key={tab.value} value={tab.value}>{tab.title}</Tabs.Trigger>
+                    )
+                })}
+            </Tabs.List>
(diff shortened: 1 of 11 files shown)


src/pages/DigitalSpecimenOverview/SubPages/DigitalSpecimenDetails.tsx
```diff
@@ -0,0 +1,87 @@
+/* Import components */
```
MelanieTheLion, line 1: Comes from DigitalSpecimenOverview, moved to its own subPage :)

src/pages/DigitalSpecimenOverview/SubPages/DigitalSpecimenDetails.scss
```diff
@@ -0,0 +1,42 @@
+@use "styles/mixins.scss" as *;
```
MelanieTheLion, line 1: Comes from DigitalSpecimenOverview.scss, became its own subPage :)

src/pages/DigitalSpecimenOverview/SubPages/DigitalSpecimenDetails.tsx
```diff
@@ -0,0 +1,87 @@
...
+        rightColumnCards = actualData; 
+    } else {
+        leftColumnCards = actualData.filter((category: MappedCategories) => LEFT_COLUMN_CATEGORIES.has(category.name));
+        rightColumnCards = actualData.filter((category: MappedCategories) => !LEFT_COLUMN_CATEGORIES.has(category.name));
+    }
+
+    return (
+        <>
+            {/* Desktop view */}
+            <main className="digital-specimen-container" id="ds-desktop-view">
```
JontySponselee-Naturalis: I assume the parentpage (since this is a subpage) also has `main`? If that is the case, it is only allowed to have one `main` in a document. You can use `section` as an alternative

src/pages/DigitalSpecimenOverview/SubPages/DigitalSpecimenDetails.tsx
```diff
@@ -0,0 +1,87 @@
...
+                            cardHeader={category.name} 
+                            fragment={category.data}
+                            AnnotateHelper={onAnnotate}
+                            {...CARD_CONFIGS[category.name as CardCategory]} 
+                        />
+                    ))}
+                </div>
+            </main>
+            {/* Mobile view */}
+            <main className="digital-specimen-container" id="ds-mobile-view">
```
JontySponselee-Naturalis: same comment as before, but you have used `display: none`  in the scss, then it is allowed to have multiple mains. But it is not that clear from the HTML structure, I think it would be better to not render the mobile part at all or use `hidden` html attribute, but then you need to have the breakpoints accessible in react as well

MelanieTheLion: I'd like to do something with this. Right now I just use breakpoint mixins and handle the mobile/desktop versions via my CSS. I will not include it in this PR, and I've adjusted my mains with sections and a proper main element on my main page. But I find the idea of globally in the app checking whether I am on mobile or not nice. Do you have an example of how you handle this in your app?

src/pages/DigitalSpecimenOverview/SubPages/DigitalSpecimenDetails.tsx
MelanieTheLion: Adjusted! You were completely right :D

src/pages/DigitalSpecimenOverview/SubPages/DigitalSpecimenDetails.tsx
JontySponselee-Naturalis: We don't have a global check as well.  We use the same approach as for the styling.

We defined it in Styles library as a custom hook. It queries the stylesheet to get the breakpoint variables.
https://gitlab.com/naturalis/bii/frontend-guild/frontend-monorepo/-/blob/main/libs/styles/src/hooks/useBreakpoints.ts?ref_type=heads

And use it as follows:
https://gitlab.com/naturalis/bii/frontend-guild/frontend-monorepo/-/blob/main/libs/styles/src/components/Layout/Header/Header.tsx?ref_type=heads#L18

src/components/Tabs/Tabs.tsx
```diff
@@ -24,15 +24,15 @@ export const DigitalSpecimenTabs = ({ defaultValue, tabs }: Props) => {
                 })}
             </Tabs.List>
 
-            <div>
+            <main>
```
JontySponselee-Naturalis: Is this correct? Isn't this component used in a page (which should have a `main`)?

MelanieTheLion: I changed it to main and sections now. The structure started out ok, but I lost it along the way.

JontySponselee-Naturalis (approved):

PR #267, opened 2026-07-30
Fix taxonomy search for annotations

MelanieTheLion:
See https://naturalis.atlassian.net/browse/DD-3202

Taxonomy tree search when adding an annotation on identification produced inconsistent results. I found out that I sometimes need to return `label` and sometimes `labelHtml`, so I added both.


southeo (approved):
nice!

PR #269, opened 2026-08-05
Added virtual collection images to the digital specimen

MelanieTheLion:
In this PR:
- Using the /full endpoint to retrieve the first jpg image of the digital specimen
- Test suite for components

[Design](https://www.figma.com/design/CW7NHzRh3Eu5555lxqNrUr/DiSSCover-v2---UX-design?node-id=1880-38764&t=UIRxt9md3HgAF8Oc-0)

[Jira link](https://naturalis.atlassian.net/browse/DD-3222)

**What it looks like:**
_On desktop:_
image

_On mobile:_
image

### DIFF
--- a/src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
+++ b/src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
@@ -59,6 +59,7 @@ export const DigitalMediaCard = ({ specimen }: Props ) => {
                 {mainImage ? (
                     <a href={`/dm/${cleanDoiPath}`}>
                         <div 
+                            role="img"
                             style={{ backgroundImage: `url(${mainImage})` }} 
                             className="digital-media-card-image"
                         ></div>
--- a/src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
+++ b/src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
@@ -60,6 +60,7 @@ export const DigitalMediaCard = ({ specimen }: Props ) => {
                     <a href={`/dm/${cleanDoiPath}`}>
                         <div 
                             role="img"
+                            aria-label={`Image of digital specimen ${activeId}`}
                             style={{ backgroundImage: `url(${mainImage})` }} 
                             className="digital-media-card-image"
                         ></div>
--- a/src/components/ImageComponents/VirtualCollectionImage/VirtualCollectionImage.test.tsx
+++ b/src/components/ImageComponents/VirtualCollectionImage/VirtualCollectionImage.test.tsx
@@ -1,0 +1,111 @@
+/* Import test dependencies */
+import { render, screen } from 'tests/test-utils';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+/* Import component & hook */
+import { VirtualCollectionImage } from './VirtualCollectionImage';
+import { useDigitalSpecimenComplete } from 'hooks/useDigitalSpecimen';
+
+/* Mock dependencies */
+vi.mock('app/Utilities', () => ({
+    RetrieveEnvVariable: vi.fn(() => 'https://doi.org/'),
+}));
+
+vi.mock('hooks/useDigitalSpecimen', () => ({
+    useDigitalSpecimenComplete: vi.fn(),
+}));
+
+/* Base mock collection */
+const baseCollection = {
+    id: 'https://doi.org/10.1234/SAMPLE-123',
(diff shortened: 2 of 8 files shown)


src/components/ImageComponents/VirtualCollectionImage/VirtualCollectionImage.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+        } as any);
+
+        const { container } = render(<VirtualCollectionImage collection={baseCollection} />);
+
+        expect(useDigitalSpecimenComplete).toHaveBeenCalledWith({
+            doi: '10.1234/SAMPLE-123',
+            enabled: true,
+        });
+
+        const imageDiv = container.querySelector('.vc-card-image-container');
```
JontySponselee-Naturalis: try to not query on classname,  classnames tend to change more often. Better to wrap the container in `figure` and query on that role, or if this div is the img then give it `role="img"` and it will also be better for A11Y

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/img_role

src/components/ImageComponents/VirtualCollectionImage/VirtualCollectionImage.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+    const { data: specimen } = useDigitalSpecimenComplete({ 
+        doi: specimenDoi,
+        enabled: hasMedia
+    });
+
+    /* Find the first jpg/jpeg media item of the digital spacimen */
+    const firstJpgItem = specimen?.digitalMedia?.find((item: any) => {
+        const format = item?.digitalMediaObject?.['dcterms:format'];
+        return format === 'image/jpeg' || format === 'image/jpg';
```
JontySponselee-Naturalis: This is fine and you can keep it like this, but you can also write it like

```ts
return ['image/jpeg', 'image/jpg'].includes(format)
```

which is a bit easier to scale if you need to add more items eventually

src/components/ImageComponents/VirtualCollectionImage/VirtualCollectionImage.test.tsx
MelanieTheLion: I used your suggestion, it is a very good one :)

src/components/ImageComponents/VirtualCollectionImage/VirtualCollectionImage.tsx
MelanieTheLion: Changed! Feels cleaner.

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
```diff
@@ -59,6 +59,7 @@ export const DigitalMediaCard = ({ specimen }: Props ) => {
                 {mainImage ? (
                     <a href={`/dm/${cleanDoiPath}`}>
                         <div 
+                            role="img"
```
JontySponselee-Naturalis, line 62: since it is not a real `img` tag there is also no `alt`, so screen readers will just scream IMAGE without context. Supply it with a `aria-label` or `aria-labelledby` to give it context.

src/components/ImageComponents/VirtualCollectionImage/VirtualCollectionImage.tsx
```diff
@@ -0,0 +1,35 @@
...
+    /* Find the first jpg/jpeg media item of the digital spacimen */
+    const firstJpgItem = specimen?.digitalMedia?.find((item: any) => {
+        const format = item?.digitalMediaObject?.['dcterms:format'];
+        return ['image/jpeg', 'image/jpg'].includes(format);
+    });
+    
+    const firstJpgUrl = firstJpgItem?.digitalMediaObject?.['ac:accessURI'];
+
+    return hasMedia && firstJpgUrl ? (
+        <div className="vc-card-image-container" role="img" style={{ backgroundImage: `url(${firstJpgUrl})` }} />
```
JontySponselee-Naturalis: Same comment  as before

MelanieTheLion: Done!

src/components/Cards/DigitalMediaCard/DigitalMediaCard.tsx
MelanieTheLion: Done!

JontySponselee-Naturalis (approved):

